PR #35, opened 2022-02-18
Throw error on incorrect cairo linting

Orland0x:
Need to throw error if the code was not correctly linted so that the linting github action will fail in that scenario. The -c param is required for this.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -42,4 +42,4 @@ jobs:
         run: pip install -r requirements.txt
 
       - name: Run linters
-        run: yarn lint
+        run: yarn check-format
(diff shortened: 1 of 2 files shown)


bonustrack (approved):

pscott (commented):
Is the return value of `cairo-format` checked somewhere ?

pscott:
I added a commit which does two things:
1. Add a `check-format` script to the `package.json` file, and renamed the `lint` script to `format`. This way developers can run `yarn format` to actually format the files, and our CI can use `check-format` to ensure correct formating.
2. Modify our github action to call `check-format` rather than `lint` :)

I've force-pushed the branch because I tested my work with two dummy commits that i removed to keep a clean history. Here's a screenshot to prove my claim https://imgur.com/a/IpznhxJ

pscott (approved):

PR #41, opened 2022-02-27
Shahak/pytest with contract deployment test

ShahakShama:


contracts/starknet/test_contract.py
```diff
@@ -0,0 +1,49 @@
+import asyncio
+import os
+import pytest
+
+from starkware.starknet.testing.contract import StarknetContract
+from starkware.starknet.testing.starknet import Starknet
+
+COMPILED_CONTRACTS_DIR = os.path.join("starknet-artifacts", "contracts", "starknet")
```
ShahakShama: I'm debating whether to take a relative path here or use `__file__` instead. I chose to go with the relative path because it won't break if we move this file into another directory, and it will work in other tests in the same way.

Do you have another suggestion on how to get starknet-artifact's path? Does yarn/hardhat define an environment variable for this?

Orland0x: Hardhat retrieves the starknet artifacts path from the param `paths[ "starknetArtifacts"]` in the config file. It currently not set and just using the default which is `starknet-artifacts`. The relative path approach you used seems fine for the python tests though.

ShahakShama: Cool, Thanks!

Orland0x:
any reason this test was placed under the `contracts/starknet` rather than `test/starknet`?

ShahakShama:
We usually put the tests next to the source code. I see the rest of the tests in this repo are in the test/ dir so I changed this one to comply with the other tests

Orland0x:
> We usually put the tests next to the source code. I see the rest of the tests in this repo are in the test/ dir so I changed this one to comply with the other tests

cool, just added the ethereum tests to an `ethereum` subdirectory

ShahakShama:
Looks good!

Orland0x:
I added a script to the `package.json` so one can now run all hardhat and python tests via `yarn test`

package.json
```diff
@@ -8,7 +8,9 @@
     "compile:l2": "hardhat starknet-compile",
     "compile": "yarn compile:l1 && yarn compile:l2",
     "deploy:l2:testnet": "hardhat starknet-deploy --starknet-network alpha",
-    "test": "hardhat test",
+    "test:hardhat": "hardhat test",
+    "test:pytest": "py.test",
+    "test": "yarn test:hardhat && yarn test:pytest",
```
ShahakShama, line 13: Note that we need to install pytest and cairo-lang from pip. Will those packages be installed in this stage?

Orland0x: yeah just fixed for the GA now

Orland0x (approved):

Orland0x (approved):
looks good

alon-dotan-starkware (approved):

PR #42, opened 2022-02-28
Add specs for milestone 1

bonustrack:
Creating this document to specify the contract interfaces and user flows that we target for milestone 1


MILESTONE-1.md
```diff
@@ -0,0 +1,101 @@
...
+- **propose(execution_hash, metadata_uri)**   
+Submit a new proposal. The execution hash is a hash from EIP-712 that include execution details, here is an example https://gist.github.com/bonustrack/45fdb2f0235d6ad3ccffa328234379aa the metadata_uri is an URL string that lead to a JSON file which include proposal metadata like title or body, this is similar than NFT tokenURI it can be "ipfs://AbC..." or "https://..." URL.
+
+### Voting strategy contract
+
+The role of the voting strategy contract is to calculate voting power for a specific user. On this milestone we just need a voting strategy that always return 1 as voting power.
+
+#### Functions
+
+- **get_vp(address, at)**
```
Orland0x: We should make the interface **get_vp(address, at, params)** so that the interface stays constant when further strategies are added. For milestone 1 the param array will be empty.

bonustrack: Ok yes just updated

MILESTONE-1.md
```diff
@@ -0,0 +1,101 @@
+# Milestone 1
+
+The first milestone include 3 contracts (in red on the diagram below), a space contract holding votes and proposals, an auth contract approving signed messages for cast vote or proposal, and a voting strategy where 1 address equal to 1 voting power. Execution, custom settings or storage proof calculations are not part of this milestone.
```
pscott: includeS*

pscott: where 1 address equalS*

MILESTONE-1.md
```diff
@@ -0,0 +1,101 @@
...
+
+![](https://user-images.githubusercontent.com/16245250/155883647-8f853411-bea9-4a73-bd69-2be850d4ffc9.png)
+
+Source: https://whimsical.com/snapshot-x-design-proposal-Gy1y3P6zTs4rdE4cStaKW4
+
+## Requirements
+
+Here is actions to cover in this milestone:
```
pscott: Here are the actions*

MILESTONE-1.md
```diff
@@ -0,0 +1,101 @@
...
+- Use auth contract as proxy
+- Create a proposal
+- Cast a vote
+- Store votes and proposals
+
+## Interfaces
+
+### Space contract
+
+This is the main contract, previously called "voting contract" the contract is responsible to store proposals and votes, it also define parameters.
```
pscott: the contract stores proposals and votes and also defines parameters* ?

MILESTONE-1.md
```diff
@@ -0,0 +1,101 @@
...
+
+- **proposals** = TBD  
+Mapping of proposals by ids with required informations: execution_hash, start, end.
+
+- **votes** = TBD  
+Mapping of votes by proposal id with required informations: address, choice, voting power.
+
+#### Functions
+
+- **vote(address, proposal, choice)**
```
ShahakShama: Could you emphasize what sort of address is this? (I'm guessing it's Ethereum).
Maybe rename this to eth_address? (TAL at [Starkgate](https://github.com/starkware-libs/starkgate-contracts/blob/02f61b33b5177d6d75c6ee20c61d22d3cd97de9f/src/starkware/starknet/apps/starkgate/cairo/token_bridge.cairo#L104) for comparison where we prefix the address with l1/l2)

bonustrack: Thanks, yes for milestone 1 it's just an Ethereum address, I've just updated the doc. Later on we will want to support both Ethereum and StarkNet addresses (with StarkNet/StarkKey authenticator).

MILESTONE-1.md
```diff
@@ -0,0 +1,101 @@
...
+- **proposals** = TBD  
+Mapping of proposals by ids with required informations: execution_hash, start, end.
+
+- **votes** = TBD  
+Mapping of votes by proposal id with required informations: address, choice, voting power.
+
+#### Functions
+
+- **vote(address, proposal, choice)**  
+Function to cast a vote, payload include the Ethereum address of the voter, the proposal id and the choice.
```
ShahakShama: Could you better explain what is choice? Is it a boolean?

MILESTONE-1.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Create a proposal
+
+This is the flow to create a proposal.
+
+- 1: User sign a proposal message on a test frontend, the message include an **execution_hash** and **metadata_uri**.
+- 2: This message and the signature are sent to the authenticator contract using **execute**, with the space contract as **target** and with the **sig**, **execution_hash** and **metadata_uri** as **data**.
+- 3: The authenticator contract accept the tx regardless of the signature being correct or not and relay that tx to the defined space.
+- 4: The space contract verify that the tx come from the authenticator contract address defined in the state variable **auth**.
+- 5: The space contract store the proposal in the variable **proposals**, this include an id defined by **next_proposal_nonce**, a **start** and **end** date defined using the current block timestamp, the space **delay** and **period**.
```
ShahakShama: > this include an id defined by **next_proposal_nonce**, a **start** and **end** date

Do you mean a chained hash of these fields? Could you specify that?

MILESTONE-1.md
Orland0x: its a value in the set {1,2,3}. Corresponding to yes, no, and abstain I believe.

pscott (commented):
quick comments

MILESTONE-1.md
Orland0x: think you are looking on the outdated file, ive made those updates

MILESTONE-1.md
Orland0x: similarly

MILESTONE-1.md
Orland0x: same situation here

MILESTONE-1.md
ShahakShama: Cool. I guess that this is trivial and we shouldn't explain it then :)
I suggest we'll add an enum for choice in cairo

MILESTONE-1.md
```diff
@@ -0,0 +1,101 @@
...
+- **get_vp(address, at, data)**  
+Returns the voting power for a specific address at a specific timestamp "at", "data" are the params required to run the voting power calculation, in milestone 1 it's just an empty array.
+
+### Auth contract
+
+The authenticator acts as a proxy to the space contract, the contract is here to validate that an address is authorized to do actions.
+
+#### Functions
+
+- **execute(target, data)**
```
pscott: In the future, the Auth contract will actually authenticate calls, correct ? If that's the case, I think we would need here:
`signature_len: felt, signature: felt*, signature_type: SignatureType` with `SignatureType` being an enum of the different supported signatures.

MILESTONE-1.md
Orland0x: get_voting_power instead of get_vp? For readability

MILESTONE-1.md
```diff
@@ -0,0 +1,101 @@
...
+A mapping of votes by proposal id with required information: address, choice, and voting power.
+
+#### Functions
+
+- **vote(address, proposal, choice)**  
+Function to cast a vote, the payload is made up of the Ethereum address of the voter, the proposal id and the choice.
+
+- **propose(address, execution_hash, metadata_uri)**   
+Submit a new proposal. The **address** is an Ethereum address, **execution_hash** is a hash from EIP-712 that include execution details, here is an example https://gist.github.com/bonustrack/45fdb2f0235d6ad3ccffa328234379aa the **metadata_uri** is an URL string that leads to a JSON file which includes proposal metadata like title or body, this is similar to NFT tokenURI. It can be "ipfs://AbC..." or "https://..." URL.
+
```
pscott: I think we need another function. Since the space contract is called by the `Authenticator`, and `Authenticator` only has a `execute` function, which cannot know if it needs to call `vote` or `propose`.
So we either need to have an entry point here that will then call `propose` or `vote`, or put it in the `authenticator`. Thoughts @bonustrack @Orland0x ?

bonustrack: The function execute need to know what method to call, this info can come from the param **data** or we must come with another way to do it. I used the same naming as DSProxy from MakerDAO where "data" are calldata see https://github.com/makerdao/dss-proxy/blob/master/src/DssProxy.sol

MILESTONE-1.md
bonustrack: What other signature type you thinking to support here? Shouldnt others type be on another Authenticator? In some case (like L1 voting) there wont be any signature.

pscott: Well I'm not sure I understand. I thought the Authenticator contract was in charge of authenticating messages, but maybe I'm wrong?

MILESTONE-1.md
pscott: Ok. So one way or the other, `data` needs to hold `function_selector` (as well as `sig, execution_hash, and metadata_uri`)

MILESTONE-1.md
```diff
@@ -0,0 +1,101 @@
...
+## Flows
+
+Here is the flows for the actions, the assumption here is all the transactions sent on StarkNet are not sent by the author of the vote or proposal itself but instead by an account that sponsor transaction fee, the relayer (https://github.com/snapshot-labs/sx-relayer).
+
+### Create a proposal
+
+This is the flow to create a proposal.
+
+- 1: User sign a proposal message on a test frontend, the message include an **execution_hash** and **metadata_uri**.
+- 2: This message and the signature are sent to the authenticator contract using **execute**, with the space contract as **target** and with the **sig**, **execution_hash** and **metadata_uri** as **data**.
```
pscott: we probably also need `function_selector` passed in here :)

MILESTONE-1.md
pscott: this loooks quite similar: https://github.com/starkware-libs/cairo-lang/blob/64a7f6aed9757d3d8d6c28bd972df73272b0cb0a/src/starkware/starknet/third_party/open_zeppelin/Account.cairo#L113-L117

MILESTONE-1.md
```diff
@@ -0,0 +1,101 @@
...
+- 3: The authenticator contract accept the tx regardless of the signature being correct or not and relay that tx to the defined space.
+- 4: The space contract verify that the tx come from the authenticator contract address defined in the state variable **auth**.
```
pscott: step 4bis should be verifying the voting_power of the proposer and making sure it's `> proposal_threshold` I think?

MILESTONE-1.md
```diff
@@ -0,0 +1,101 @@
...
+The authenticator contract address that is used to verify the addresses of voters and proposers.
+
+- **next_proposal_nonce** = 1  
+An increasing nonce to make the proposal id unique.
+
+- **proposals** = TBD  
+A mapping of proposals by ids with required information: execution_hash, start, end.
+
+- **votes** = TBD  
+A mapping of votes by proposal id with required information: address, choice, and voting power.
```
ShahakShama: I think we can avoid this mapping and save storage_vars by doing the following:

1. Adding a mapping called `has_voted` that maps between a pair of voting_address, proposal_id into a flag that says whether the address already voted on the proposal
2. Adding three mappings that map between proposal_id to a counter. One for the sum of "Yes" votes. One for the sum of "No: votes, and one for the sum of "Abstain" votes

This way we have only two writes for each vote. Instead of 3 here and at least one more to manage the array of votes per proposal.

MILESTONE-1.md
bonustrack: @ShahakShama These fields i believe should be stored in plain cuz we will need it when someone cast a vote to make sure it's done within the proposal period. We will probably want to store a hash of the voting strategies setup within the proposal to make sure strategies don't change while proposal is still going on, but we can do that on another milestone.

MILESTONE-1.md
bonustrack: There will be many authenticator, the vanilla authenticator would just allow anyone, but we will have an auth for L1 voting, another for StarkKey, another for Ethereum signature. Not just a single auth.

MILESTONE-1.md
pscott: The only problem I see is that we can't query any storage var that would show us the list of voters for a specific `proposal_id`. Don't know if that's an issue or not :) (maybe @bonustrack ? :)

pscott: Also @ShahakShama , do you think we could get away with `votes(proposal_id: felt, choice: felt) -> (count: felt)` ?

This would remove a big `if else if else if else` chunk of ugly code :p

pscott: for reference: I implemented it in another branch if we ever decide to switch to it:

https://github.com/snapshot-labs/sx-core/commit/7b2060fe034945aa98ea351fe21f4cce7ed229bb

bonustrack: > The only problem I see is that we can't query any storage var that would show us the list of voters for a specific `proposal_id`. Don't know if that's an issue or not :) (maybe @bonustrack ? :)

@psott It won't be an issue we can emit an event and get the information with the API

ShahakShama: @pscott Great Idea!

MILESTONE-1.md
```diff
@@ -40,11 +40,11 @@ The authenticator contract address that is used to verify the addresses of voter
 - **next_proposal_nonce** = 1  
 An increasing nonce to make the proposal id unique.
 
-- **proposals** = TBD  
-A mapping of proposals by ids with required information: execution_hash, start, end.
+- **proposals** = { 1: { start: 123.., end: 456.., execution_hash: '0x123...', scores: { 0: 123, 1: 45, 3: 6 }]
```
ShahakShama: Having a list or a mapping inside a storage_var's value is complicated. This is because you can't simply have a pointer in there because the segment it points to won't be stored for future calls to this contract
The better way to do this here is to have two mappings:
1. A map between a proposal_id to a struct containing start, end and execution_hash. 
2. A map between a proposal_id and choice to number of votes

MILESTONE-1.md
```diff
@@ -40,11 +40,11 @@ The authenticator contract address that is used to verify the addresses of voter
...
+A mapping of proposals by ids with required information: execution_hash, start, end, scores with a counter for the vote For, Against and Abstain.
 
-- **votes** = TBD  
-A mapping of votes by proposal id with required information: address, choice, and voting power.
+- **has_voted** = { 1: [0x123..., 0x456...], 2: [0x789...] }
```
ShahakShama: Same as above, this map should be between proposal_id and vote_address to a boolean

MILESTONE-1.md
bonustrack: I see thanks, I'll update the doc

Orland0x (approved):

bonustrack:
This milestone is done, I'm merging it for now we will have another doc with milestone 2

PR #44, opened 2022-02-28
add Choice enum

ShahakShama:


contracts/starknet/objects/choice.cairo
```diff
@@ -0,0 +1,4 @@
+namespace Choice:
+    const YES = 0
```
bonustrack: If we can use 0, I suggest we simply use the same format from OpenZeppelin Governor which is:

Against = 0
For = 1
Abstain = 2

See:
https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/governance/extensions/GovernorCountingSimple.sol#L18-L20

bonustrack (approved):

Orland0x:
I was under the impression that we couldn't use 0 for one of the choices as it was the default value. Is this not an issue here?

ShahakShama:
I initially thought this won't be a problem because we don't have a storage var for choice only, but it's better this way anyway just to be safe :)

Orland0x (approved):

PR #45, opened 2022-03-01
Vanilla authenticator

pscott:
As per what was described in [milestone1](https://github.com/snapshot-labs/sx-core/pull/42/files/56e5638697ad7360f2e76265b4d74d05a41b40b6#diff-d404420bd39a9b8f0b2fd88633cb896d9ad8c7855d74de505fa51dffa35accbb), here is a vanilla authenticator contract that does not check any signature and only forwards the call.


Orland0x:
looks good

contracts/starknet/authenticator/authenticator.cairo
```diff
@@ -0,0 +1,15 @@
+%lang starknet
+from starkware.starknet.common.syscalls import call_contract
+
+# Forwards `data` to `target` without verifying anything.
+# TODO: use ADDRESS instead of felt
```
Orland0x: ive got an address type implemented so we can change soon

Orland0x (approved):

contracts/starknet/authenticator/authenticator.cairo
```diff
@@ -0,0 +1,15 @@
...
+@external
+func execute{syscall_ptr: felt*, range_check_ptr}(to: felt, selector: felt, calldata_len: felt, calldata: felt*):
+    # TODO: Actually verify the signature
+
+    # Call the space contract
```
bonustrack: We should not mention "space" in this contract, this contract will be general purpose, it could be used to call other kind of contract.

contracts/starknet/authenticator/authenticator.cairo
bonustrack: What the use of Address type here? Isn't Address type used only for L1 addresses?

pscott: We could use L1Address and StarknetAddress :) I'm in favor of replacing as many `felt` as possible, because otherwise we end up with a code that is just full of `felt`s :p

pscott: Currently since there's no way to actually wrap a type (other than creating a struct), I'm not sure it makes sense to actually use `StarknetAddress`. Therefore I'm removing this comment :)

pscott:
Tests are failing because we need to define the `SPACE_CONTRACT` if we wish to interact with it :p But at least we've got the code required to interact with the `authenticator` contracts (useful for @bonustrack :))

pscott:
I suggest merging this bad boy (I commented out the tests for the vanilla auth), so that I can then merge `develop` into the other branch (`vanilla_space`). Since both branches need each other, I think it's the only way :) (or we could directly merge this branch in `vanilla_space`) but I think it's cleaner to merge this one into `develop` first. @bonustrack wdyt? :)

bonustrack (approved):
Sound good!

PR #47, opened 2022-03-01
Vanilla Space Contract

pscott:
As per what was described in [milestone1](https://github.com/snapshot-labs/sx-core/pull/42/files/56e5638697ad7360f2e76265b4d74d05a41b40b6#diff-d404420bd39a9b8f0b2fd88633cb896d9ad8c7855d74de505fa51dffa35accbb), here is a vanilla space contract (missing correct types and tests).


contracts/starknet/space/space.cairo
```diff
@@ -0,0 +1,134 @@
...
+
+    return ()
+end
+
+# TODO: should be address not felt, and choice should be of enum Choice not felt
+@external
+func vote{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr: felt}(eth_address: felt, proposal_id: felt, choice: felt) -> ():
+
+    # Verify that the caller is the authenticator contract.
+    let (caller_address) = get_caller_address()
```
ShahakShama: Consider defining a function called authenticator_only that checks the caller is the authenticator. Something that looks like a modifier in solidity. We do something similar in [Starkgate](https://github.com/starkware-libs/starkgate-contracts/blob/02f61b33b5177d6d75c6ee20c61d22d3cd97de9f/src/starkware/starknet/apps/starkgate/cairo/ERC20.cairo#L107)

contracts/starknet/space/space.cairo
```diff
@@ -0,0 +1,134 @@
...
+@storage_var
+func next_proposal_nonce() -> (nonce : felt):
+end
+
+@storage_var
+func proposals(proposal_id: felt) -> (proposal : Proposal):
+end
+
+@storage_var
+func votes(proposal_id: felt) -> (vote : Vote):
```
ShahakShama: I think this should map into an array of votes and not a single vote. If so, we should make the following changes
1. change the type to Vote*
2. add a storage_var called n_votes
3. call alloc in the constructor and write the return value into votes

contracts/starknet/space/space.cairo
```diff
@@ -0,0 +1,134 @@
...
+
+struct Proposal:
+    member execution_hash: felt # TODO: Use Hash type
+    member start_block: felt
+    member end_block: felt
+end
+
+# TODO: use L1Address instead of felt
+@contract_interface
+namespace IVotingStrategy:
```
ShahakShama: Put this in a different file (contracts/starknet/strategies/IVotingStrategy.cairo)

contracts/starknet/space/space.cairo
```diff
@@ -0,0 +1,134 @@
+%lang starknet
+
+from starkware.starknet.common.syscalls import ( get_caller_address, get_block_number )
+from starkware.cairo.common.cairo_builtins import HashBuiltin
+from starkware.cairo.common.math import assert_lt, assert_le, assert_nn, assert_not_zero
+
+struct Vote:
```
ShahakShama: Implement these structs in a different file so that we can use them in other contracts

contracts/starknet/space/space.cairo
```diff
@@ -0,0 +1,134 @@
...
+    member choice: felt # TODO use Choice enum
```
ShahakShama: There's no need for this TODO as we don't have enum in Cairo.
What we do instead is that we create constants of the numbers of the enum, and put all these constants under the same namespace.
Therefore, we won't have a choice type and this will remain felt

contracts/starknet/space/space.cairo
ShahakShama: Actually, this doesn't work as the data in the Vote* itself isn't accessible
We need to have some sort of vote_id or something

contracts/starknet/space/space.cairo
pscott: good idea :)
[309caaa](https://github.com/snapshot-labs/sx-core/pull/47/commits/309caaa37d3b57d01e5e7b81344476e4e06a8945)

contracts/starknet/space/space.cairo
pscott: https://github.com/snapshot-labs/sx-core/pull/47/commits/cfedfb08beec24e956cf59f855a2118e54051567 :D

contracts/starknet/space/space.cairo
pscott: I was referring to your Choice implementation here:
https://github.com/snapshot-labs/sx-core/commit/8f760dac692479e52bdfc26eda77c33b5a9ace9a

I think we should use this enum here in the Voting structure. What do you think about it ? :)

pscott: Ok my bad, I thought we could use namespace as `types` but it looks like we can't :) I guess I'm just used to Rust and C :)

Fix: [Remove comments asking to use Choice as a type (impossible)](https://github.com/snapshot-labs/sx-core/pull/47/commits/e120f74a7b050abc3fd24e9055a006d92f0828f2)

contracts/starknet/space/space.cairo
pscott: https://github.com/snapshot-labs/sx-core/pull/47/commits/cfedfb08beec24e956cf59f855a2118e54051567 :D

contracts/starknet/space/space.cairo
pscott: Do you know if there's a way for us to check if the `choice` argument is one of the `Choice` possible value ? Other than doing some `if .... else if ... else if...` for every possible value ? (I'd like to use it in the `vote` function, right before inserting the vote in the storage variable, to make sure the vote is valid)

pscott:
Tests are currently not working (because state is not updating... don't know why) :) 

EDIT: fixed by using `invoke` instead of `call`...

contracts/starknet/space/space.cairo
```diff
@@ -0,0 +1,250 @@
...
+                tempvar range_check_ptr = range_check_ptr
+                tempvar syscall_ptr = syscall_ptr
+                tempvar pedersen_ptr = pedersen_ptr
+                tempvar voting_power = voting_power
+            else:
+                # choice is not a valid choice ?!
+                return ()
+            end
+        end
+    end
```
pscott: We might be able to remove this bad boy if we use this: https://github.com/snapshot-labs/sx-core/pull/42#discussion_r818913667

(and simply  check that `Choice.Against <= choice <= Choice.Abstain`)

pscott: (implemented on this branch if you want to 👀 it : https://github.com/snapshot-labs/sx-core/commit/7b2060fe034945aa98ea351fe21f4cce7ed229bb)

contracts/starknet/objects/proposal.cairo
```diff
@@ -0,0 +1,6 @@
+struct Proposal:
+    member execution_hash : felt  # TODO: Use Hash type
+    member metadata_uri : felt  # TODO: Use String type
+    member start_block : felt
```
bonustrack: Did you tried with timestamp here?

test/starknet/vanilla_space.ts
```diff
@@ -0,0 +1,134 @@
...
+    // -- Casts a vote FOR --
+    {
+      const voter_address = proposer_address;
+      await vanillaAuthenticator.invoke(EXECUTE_METHOD, {
+        to: space_contract,
+        function_selector: BigInt(getSelectorFromName(VOTE_METHOD)),
+        calldata: [voter_address, proposal_id, FOR],
+      });
+
+      const { proposal_info: proposal_info } = await vanillaSpace.call('get_proposal_info', {
```
bonustrack: This works too
```suggestion
      const { proposal_info } = await vanillaSpace.call('get_proposal_info', {
```

contracts/starknet/space/space.cairo
```diff
@@ -0,0 +1,256 @@
...
+        # Not enough voting power to create a proposal
+        assert 1 = 0
+    end
+
+    # Create the proposal and its proposal id
+    let proposal = Proposal(execution_hash, metadata_uri, start_block, end_block)
+    let (proposal_id) = next_proposal_nonce.read()
+
+    # Store the proposal
+    proposal_registry.write(proposal_id, proposal)
```
bonustrack, line 193: Can you emit event here with all the info + proposer_address

contracts/starknet/space/space.cairo
bonustrack: If it work i'm all for it, is there a way to compare the number of steps and amount of storage required?

contracts/starknet/space/space.cairo
```diff
@@ -0,0 +1,256 @@
...
+                tempvar voting_power = voting_power
+            else:
+                # choice is not a valid choice ?!
+                return ()
+            end
+        end
+    end
+
+    let vote = Vote(choice=choice, voting_power=voting_power)
+    vote_registry.write(proposal_id, voter_address, vote)
```
bonustrack, line 144: Can you also emit event here?

test/starknet/shared/types.ts
```diff
@@ -1,5 +1,9 @@
 import { assert, hexToBytes, bytesToHex } from './helpers';
 
+export const AGAINST = BigInt(1);
```
bonustrack: If we can't use 0 for "Against", it's better to have the choice in the order it will be render on UI like this:
1: For
2: Against
3: Abstain

contracts/starknet/space/space.cairo
```diff
@@ -0,0 +1,256 @@
...
+
+@view
+func get_nonce{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr : felt}() -> (
+        next_nonce : felt):
+    let (nonce) = next_proposal_nonce.read()
+    return (next_nonce=nonce)
+end
+
+@view
+func get_delay{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr : felt}() -> (
```
bonustrack: If it's just for UI we may not need these views functions, we would just need to emit events with the space settings and the API would store this information.

test/starknet/shared/types.ts
pscott: Ok! Was following the order of `choice.cairo`. I've updated everything at the same time in https://github.com/snapshot-labs/sx-core/pull/47/commits/355c2f92425bdcdc43a4d6d6b76b91e8b485e373 :)

contracts/starknet/space/space.cairo
pscott: Yeah I initially wrote those with testing in mind. I've removed everything except `get_vote_info` and `get_proposal_info` in [2a1fde7](https://github.com/snapshot-labs/sx-core/pull/47/commits/2a1fde787effff41c2b618dd0c058d31e57020c4) :)

contracts/starknet/space/space.cairo
pscott: [5997c7b](https://github.com/snapshot-labs/sx-core/pull/47/commits/5997c7b45cce5cac00f09c5bd84fe6fc4b142120) :D

contracts/starknet/space/space.cairo
pscott: [5997c7b](https://github.com/snapshot-labs/sx-core/pull/47/commits/5997c7b45cce5cac00f09c5bd84fe6fc4b142120) :D

contracts/starknet/objects/proposal.cairo
pscott: https://github.com/snapshot-labs/sx-core/pull/47/commits/ce41532f7ea1496460fdde71f80cb5e24d558a25

test/starknet/vanilla_space.ts
```diff
@@ -0,0 +1,134 @@
...
+
+      // We can't directly compare the `info` object because we don't know for sure the value of `start_block` (and hence `end_block`),
+      // so we compare it element by element (except start_block and end_block for which we simply compare their difference to `VOTING_PERIOD`).
+      expect(proposal_info.proposal.execution_hash).to.deep.equal(execution_hash);
+      expect(proposal_info.proposal.metadata_uri).to.deep.equal(metadata_uri);
+      expect(
+        proposal_info.proposal.end_timestamp - proposal_info.proposal.start_timestamp
+      ).to.deep.equal(VOTING_PERIOD);
+
+      const _for = SplitUint256.fromObj(proposal_info.power_for).toUint();
```
pscott, line 85: I introduced `fromObj` because `proposal_info.power_for` is of type `{low: bigint; high: bigint}` when I would like a `SplitUin256{low: bigint, high: bigint}`. I tried casting by doing `const _for = (proposal_info.power_for as SplitUint256).toUint();` but didn't manage to make it work. Maybe the `.ts` god @bonustrack could help on that :D

contracts/starknet/space/space.cairo
pscott: I get the exact same number: `518` steps :)

pscott: I get the exact same number: `518` steps :)

contracts/starknet/objects/vote.cairo
```diff
@@ -0,0 +1,9 @@
+%lang starknet
+
+from contracts.starknet.lib.types import EthAddress
```
Orland0x: We should consolidate the types either into the types.cairo file or the objects directory. I wonder if there is more overhead doing it one way rather than the other.

contracts/starknet/space/space.cairo
```diff
@@ -0,0 +1,214 @@
...
+    assert_le(proposal.start_timestamp, current_timestamp)
+
+    # Make sure voter has not already voted
+    let (prev_vote) = vote_registry.read(proposal_id, voter_address)
+    if prev_vote.choice != 0:
+        # Voter has already voted!
+        assert 1 = 0
+    end
+
+    let (local params : felt*) = alloc()
```
Orland0x: We should pass these params as an argument for `vote`, so that the interface doesnt change once parameters are needed. Will just be an empty array for milestone 1.

contracts/starknet/space/space.cairo
```diff
@@ -0,0 +1,214 @@
...
+    let (strategy_contract) = voting_strategy.read()
+
+    let (user_voting_power) = IVotingStrategy.get_voting_power(
+        contract_address=strategy_contract,
+        timestamp=current_timestamp,
```
Orland0x, line 124: This will effectively set the snapshot at `current_timestamp`, which will be different for each vote. Dont we want to fix it?

bonustrack: Good point, this should be the proposal "start" instead

contracts/starknet/space/space.cairo
```diff
@@ -0,0 +1,214 @@
...
+        params_len=0,
+        params=params)
+
+    # Make sure `choice` is a valid choice
+    assert_le(Choice.FOR, choice)
+    assert_le(choice, Choice.ABSTAIN)
+
+    let (previous_voting_power) = vote_power.read(proposal_id, choice)
+    # TODO: decide what to do with carry
+    let (new_voting_power, carry) = uint256_add(user_voting_power, previous_voting_power)
```
Orland0x, line 134: We should throw an assertion error if the carry is non zero. It is then up to the strategy creators to create strategies that yield voting powers that wont violate this.

contracts/starknet/space/space.cairo
pscott: [Use params in execute, vote, and proposal](https://github.com/snapshot-labs/sx-core/pull/47/commits/a3a55bd898bbf1061ecd8e15041e4ca7a6a6ed80) :D

contracts/starknet/space/space.cairo
pscott: [Throw if overflow happens](https://github.com/snapshot-labs/sx-core/pull/47/commits/7e8bc4562c966e2e46807f2210859a4e565158e7)

contracts/starknet/space/space.cairo
pscott: [Add ethereum block number to proposal struct](https://github.com/snapshot-labs/sx-core/pull/47/commits/b886d487560f1eea040dcb397f16e1eb08ee3e11) :D

contracts/starknet/objects/vote.cairo
pscott: @ShahakShama do you have any preference between a single `types` file and a `obj/` folder with files for each type ?

ShahakShama: Both are good

pscott:
Decided to use a `lib/` folder with a single file per new struct / type. https://github.com/snapshot-labs/sx-core/pull/47/commits/8371c1e6c472ed4b00a93235f59bccce0eb4471d

Can merge as soon as I get two approvals :)

Orland0x:
looks good

Orland0x (approved):

bonustrack (approved):
Looks great!

PR #55, opened 2022-03-11
Single slot proof strategy

Orland0x:
- Single Slot proof voting strategy and test
- removed the demo python test from starkware and adjusted the package.json accordingly.


contracts/starknet/account/Account.cairo
```diff
@@ -0,0 +1,190 @@
+%lang starknet
```
pscott, line 1: Is this the `Account` from OZ? Should we mention it somewhere ?

contracts/starknet/strategies/single_slot_proof.cairo
```diff
@@ -0,0 +1,92 @@
...
+namespace IFactsRegistry:
+    func get_storage(
+            block : felt, account_160 : felt, slot : StorageSlot, proof_sizes_bytes_len : felt,
+            proof_sizes_bytes : felt*, proof_sizes_words_len : felt, proof_sizes_words : felt*,
+            proofs_concat_len : felt, proofs_concat : felt*) -> (
+            res_bytes_len : felt, res_len : felt, res : felt*):
+    end
+end
+
+# Address of the fact registry. This is an immutable value that can be set at contract deployment only.
```
pscott, line 18: In a future version, should this be updatable (in case of a re-deployment of the fossil contract?

contracts/starknet/strategies/single_slot_proof.cairo
```diff
@@ -0,0 +1,92 @@
...
+        proofs_concat_len,
+        proofs_concat)
+
+    # Converting the returned value to a cairo uint256
+    let (voting_power) = ints_to_uint256(storage_bytes_len, storage_len, storage)
+
+    return (voting_power)
+end
+
+@view
```
pscott, line 59: Does this function need to be @view ? IIUC, view means this will be `callable` as a read only function.. but if it's only used  in `get_voting_power`, I don't think we need @view on it ? (this applied to the other `@views` functions :p)

contracts/starknet/strategies/single_slot_proof.cairo
```diff
@@ -0,0 +1,92 @@
...
+func decode_param_array{range_check_ptr}(param_array_len : felt, param_array : felt*) -> (
+        slot : StorageSlot, proof_sizes_bytes_len : felt, proof_sizes_bytes : felt*,
+        proof_sizes_words_len : felt, proof_sizes_words : felt*, proofs_concat_len : felt,
+        proofs_concat : felt*):
+    assert_nn_le(4, param_array_len)
+    let slot : StorageSlot = StorageSlot(
+        param_array[0], param_array[1], param_array[2], param_array[3])
+    let num_nodes = param_array[4]
+    let proof_sizes_bytes_len = num_nodes
+    let proof_sizes_bytes = param_array + 5
```
pscott, line 69: is this equivalent to `param_array[5]` ?

contracts/starknet/strategies/single_slot_proof.cairo
```diff
@@ -0,0 +1,92 @@
...
+
+    return (voting_power)
+end
+
+@view
+func decode_param_array{range_check_ptr}(param_array_len : felt, param_array : felt*) -> (
+        slot : StorageSlot, proof_sizes_bytes_len : felt, proof_sizes_bytes : felt*,
+        proof_sizes_words_len : felt, proof_sizes_words : felt*, proofs_concat_len : felt,
+        proofs_concat : felt*):
+    assert_nn_le(4, param_array_len)
```
pscott, line 64: Are we sure `4` is enough ? It looks to me like we're at least checking `param_array[4]` (which is 5 and not 4) for `num_nodes`?

hardhat.config.ts
```diff
@@ -35,13 +35,26 @@ task('accounts', 'Prints the list of accounts', async (args, hre) => {
...
+        settings: {
+          optimizer: {
+            enabled: true,
+            runs: 10,
+          },
+        },
       },
-    },
+      {
+        version: '0.8.9',
```
pscott, line 40: two different compilers ? :)

pscott (commented):
Have not reviewed tests yet :) Nice work, way more code then I was expecting but I don't think we can get out with less than that :(

contracts/starknet/strategies/single_slot_proof.cairo
Orland0x: that sounds like a bad idea to me, this is a library contract after all and will have no 'owner' so who would have the permissions to change it? In the case of fossil being redeployed we would then redeploy the strategies that depend on it.

contracts/starknet/account/Account.cairo
Orland0x: yes its OZ. Yeah, where do you think would be best, maybe the readme? We could also add the OZ repo as a submodule, might be better idk

hardhat.config.ts
Orland0x: I wrote the zodiac module in 8.6 as that is what the rest of the gnosis zodiac stuff was written with, then fossil uses 8.9. We may want to submit a PR with the zodiac module to the main zodiac repo at some point too so may be best to keep that 8.6.

Orland0x:
> Have not reviewed tests yet :) Nice work, way more code then I was expecting but I don't think we can get out with less than that :(

yeah ik its mostly the data type conversions both on and off chain which are long and a bit of a PITA. But yeah dont see much way around it for now, ive being lobbying marcello a bit to integrate uint256 into Fossil (then we wont need the ints sequence to uint256 stuff) but we'll see whether that happens

contracts/starknet/strategies/single_slot_proof.cairo
Orland0x: oh okay wasnt aware of this. just thought `@view` was all non state changing functions. Will amend

contracts/starknet/strategies/single_slot_proof.cairo
Orland0x: so the min length is 5. and that checks 5 > 4 right? its a strict inequality i think

contracts/starknet/strategies/single_slot_proof.cairo
Orland0x: `proof_sizes_bytes` is a pointer so no

PR #65, opened 2022-03-17
Milestone 2

Orland0x:
Spec for milestone 2 

Need to add spec for Zodiac execution strategy and corresponding flow.


docs/milestones/2.md
```diff
@@ -0,0 +1,90 @@
...
+
+### StarkNet Commit Contract
+
+This global contract will allow an alternate way to interact with Snapshot X: Via an L1 transaction. The core benefit of this feature is to allow one to interact with Snapshot X without a signature. This will give smart contract accounts such as multi-sig wallets the ability to participate. 
+
+The contract will have a single storage variable: The address of the global L1 Auth contract on StarkNet. This will be an immutable variable set at construction of the contract.
+
+#### Functions 
+
+- **submitCommitment(commitment: bytes32)** \
```
bonustrack: I would just call it `commit` like this
```suggestion
- **commit(hash: bytes32)** \
```

docs/milestones/2.md
```diff
@@ -0,0 +1,90 @@
...
+Sends the hash provided to the StarkNet messaging bridge with the L1 Auth contract as the destination address.
+
+### L1 Auth Contract 
+
+This global contract will receive the hash of the parameters required for the relevant interaction parameters and will hash them in the contract to check that they match a hash submitted via the StarkNet Commit contract. 
+
+#### Functions 
+
+- **propose_via_L1(space: felt, execution_hash: felt array, metadata_uri: felt array, params: felt array)** \
```
bonustrack: The Auth L1 tx contract should be general purpose, something very similar to vanilla auth with a `execute` function that take arbitrary call.

docs/milestones/2.md
```diff
@@ -0,0 +1,89 @@
...
+
+[Source](https://whimsical.com/snapshot-x-milestone-2-7UJARrFFC3LbnZexGEtQQB)
+
+### Single Slot Proof Voting Strategy Contract
+
+We add the principal voting strategy of Snapshot X to this milestone. The single slot proof strategy will allow the balances of the majority [2] of ERC20 and ERC721 tokens to be used to find voting power. 
+
+#### Functions
+
+- **get_voting_power(address: EthAddress, ethereum_block_number: felt, params: felt array)**
```
pscott, line 25: will this be `ethereum_block_number` or a `timestamp` ?

docs/milestones/2.md
```diff
@@ -0,0 +1,89 @@
...
+### Single Slot Proof Voting Strategy Contract
+
+We add the principal voting strategy of Snapshot X to this milestone. The single slot proof strategy will allow the balances of the majority [2] of ERC20 and ERC721 tokens to be used to find voting power. 
+
+#### Functions
+
+- **get_voting_power(address: EthAddress, ethereum_block_number: felt, params: felt array)**  
+Returns the voting power for a specific address at a specific Ethereum block number **block_number**, **params** are the storage proof data required by the Fossil module to prove the contents of a slot. In the case of finding the voting power from an ERC20 balance, the slot will correspond to the storage slot in the ERC20 contract where the balance of **address** is stored. 
+
+### StarkNet Commit Contract
```
pscott, line 28: `Starknet Commit Contract` sounds like it something "bigger" than Snapshot X. As if anyone on Starknet could benefit from it?

docs/milestones/2.md
```diff
@@ -0,0 +1,89 @@
...
+
+Auth contract for interactions with Snapshot X via L1. 
+
+#### Functions 
+
+- **commit_handler(hash: felt array)** \
+L1 handler function that receives the hash sent by the Starknet Commit contract and stores it in its state.
+
+- **execute(to : felt, function_selector : felt, calldata : felt array)** \
+General purpose athentication for all types of interaction where the interaction parameters are sent and hashed to check that they match a hash recieved from L1. If so, the function given by **function_selector** at the **to** address is called with **calldata** as parameters.
```
pscott: authentication*

docs/milestones/2.md
```diff
@@ -0,0 +1,89 @@
...
+- 7a: The proposal creation flow then continues just as specified in milestone 1 from point 4 onwards. 
+
+#### Cast a vote via L1
+- 4a: User submits a StarkNet transaction to the **vote_via_L1** function of L1 Auth contract with the vote parameters as arguments: The address of the space the user wants to create a proposal in **space**, the ID of the proposal **proposal_id**, the voter's choice **choice**, and the parameters required to get the voting power of the user **params** (This will most commonly be an encoded storage proof).
+- 5a: The contract computes the hash of the parameters and checks that it matches a hash received via the bridge (in step 3). 
+- 6a: The contract then submits an internal transaction to the **vote** function of the of the space contract specified by **space** with the **proposal_id**, the **choice**, and the **params** as arguments. 
+- 7a: The vote submission flow then continues just as specified in milestone 1 from point 4 onwards.
```
pscott: 4b, 5b, 6b, 7b?

docs/milestones/2.md
```diff
@@ -0,0 +1,89 @@
...
+- 2: The contract then submits a transaction to StarkNet over the messaging bridge with the destination address immutably set in the contract as the L1 Tx Authenticator, and the payload as the hash supplied.
+- 3: The L1 Tx Authenticator receives the message and stores the payload parameters, it now waits for the strategy parameters to be submitted.
+
+#### Propose via L1 
+- 4a: User submits a StarkNet transaction to the **propose_via_L1** function of L1 Auth contract with the proposal parameters as arguments: The address of the space the user wants to create a proposal in **space**, the hash of the execution details **execution_hash**, the URI of the metadata **metadata_uri**, and the parameters required to get the voting power of the user **params** (This will most commonly be an encoded storage proof).
+- 5a: The contract computes the hash of the parameters and checks that it matches a hash received via the bridge (in step 3). 
+- 6a: The contract then submits an internal transaction to the **propose** function of the of the space contract specified by **space** with the **execution_hash**, the **metadata_uri**, and the **params** as arguments.
```
pscott, line 86: of the space*

docs/milestones/2.md
```diff
@@ -0,0 +1,89 @@
...
+
+A way to execute Gnosis Safe transactions that are included in a proposal, once the vote has concluded successfully.
+
+#### Functions
+
+**receiveProposal(executionDetails: Uint256, hasPassed: Uint256, txHashes: bytes32[])** \
+Consumes message from StarkNet bridge, checks proposal has passed, and then hashes the **txHashes** array to check it is equal to **executionDetails**. If so, it stores the **txHashes** ready to be executed in a following transaction.
+
+**executeProposalTxBatch(proposalIndex: Uint256, Raw Tx parameter arrays ...)** \
+Hashes each transaction in turn and checks that it is equal to the corresponding txHash stored. If correct, executes that transaction.
```
pscott: what's the use of `proposalIndex` ? I don't think the function `receiveProposal` mentions it, so is it a variable that's internal to the Zodiac Module Contract? It's not the same `proposalIndex` as the one on the L2 space contract right?

docs/milestones/2.md
bonustrack: We can keep using a block number for now, if we want a timestamp it will require more work we can do that on another milestone

bonustrack (approved):

docs/milestones/2.md
Orland0x: yeah me and fabien discussed this on discord, could be used by anyone who wants to commit data to L2 via an L1 tx.

docs/milestones/2.md
Orland0x: Yeah its just a counter that is incremented each time a new proposal is received, its used to access the proposals in the contract so works like an ID. I called it `proposalIndex` rather than `proposalID` so that it was not mixed up with the L2 one. maybe we should use a different name all together for clarity. `proposalNumber`?

pscott (approved):

PR #67, opened 2022-03-28
add Execution

pscott:
This PR adds the `execution` feature.
Basically, this PR adds a `finalize_proposal` method to the `space` contract, and adds a `zodiac_relayer.cairo` contract that will relay the execution to L1. L1 smart-contract is also modified to have a mapping of whitelisted Spaces that can trigger the execution.

This `finalize_proposal` can be called by anyone, but only after the `proposal_period` has ended.
It sends a message to the execution contract stored in `executor`, and notes the proposal as `executed`. It is up to the execution contract to decide what to do with the information provided.


pscott:
This PR is finally ready!
The tests are failing, but [they are failing on another PR too...](https://github.com/snapshot-labs/sx-core/runs/5731963243?check_suite_focus=true) (for some reason it looks like starknet-devnet can't connect to port 8000? will have to check)

contracts/starknet/space/space.cairo
```diff
@@ -75,25 +85,36 @@ end
...
     voting_strategy.write(_voting_strategy)
     authenticator.write(_authenticator)
+    l1_executor.write(EthAddress(_l1_executor))
     next_proposal_nonce.write(1)
 
     return ()
 end
 
+# TODO: L1 needs to know about L2 address, but L2 needs to know about the L2 address... need to fix that.
+# TODO: this should either be on the l1 contract or the l2 contract. Since l1 contract has an `owner` I think it should be on l1.
```
pscott: This needs to be addressed (we can merge like this, but will definitely need to be addressed soon)

test/starknet/shared/helpers.ts
```diff
@@ -17,3 +32,61 @@ export function bytesToHex(bytes: number[]): string {
...
+  console.log('Deploying auth...');
+  const vanillaAuthenticator = (await vanillaAuthenticatorFactory.deploy()) as StarknetContract;
+  console.log('Deploying strat...');
+  const vanillaVotingStrategy = (await vanillaVotingStategyFactory.deploy()) as StarknetContract;
+  const voting_strategy = BigInt(vanillaVotingStrategy.address);
+  const authenticator = BigInt(vanillaAuthenticator.address);
+  console.log('Deploying space...');
+
+  // This should be declared along with the other const but doing so will make the compiler unhappy as `SplitUin256`
+  // will be undefined for some reason?
```
pscott: I need some TS gods to help me here... initially had it along with the other consts but it simply didn't work ^^

pscott:
Currently working on adding:

- [x] An interface for the execution
- [x] storing the execution parameters (as a Hash) and recovering it
- [x] A `zodiac_l1` execution strategy
- [x] Adding a check on the l1 zodiac module to add a whitelist of accepted callers

contracts/ethereum/SnapshotXZodiacModule/ProposalRelayer.sol
```diff
@@ -15,64 +15,73 @@ contract SnapshotXProposalRelayer is Guardable {
...
 
   /**
    * @dev Checks whether proposal has been received on L1 yet
-   * @param executionDetails Hash of all the transactions in the proposal
+   * @param executionHash Hash of all the transactions in the proposal
    * @param hasPassed Whether the proposal passed
    * @return isReceived Has the proposal been received
    */
-  function isFinalizedProposalRecieved(uint256 executionDetails, uint256 hasPassed)
+  function isFinalizedProposalReceived(uint256 executionHash, uint256 hasPassed)
```
Orland0x: This needs to be updated to the new message payload format. Tbh we dont rly need this function right? I just added it for the sake of completeness but we could remove

contracts/ethereum/SnapshotXZodiacModule/SnapshotXL1Executor.sol
```diff
@@ -53,15 +56,15 @@ contract SnapshotXL1Executor is Module, SnapshotXProposalRelayer {
    * @param _owner Address of the owner of this contract
    * @param _avatar Address that will ultimately execute function calls
    * @param _target Address that this contract will pass transactions to
-   * @param _decisionExecutorL2 Address of the StarkNet contract that will send execution details to this contract in a L2 -> L1 message
+   * @param _l2ExecutionRelayer Address of the StarkNet contract that will send execution details to this contract in a L2 -> L1 message
    * @param _starknetCore Address of the StarkNet Core contract
    */
   event SnapshotXL1ExecutorSetUpComplete(
```
Orland0x, line 62: should we add the whitelisted spaces to this event?

contracts/starknet/space/space.cairo
```diff
@@ -24,13 +29,16 @@ end
...
+func voting_strategies(voting_strategy_contract : felt) -> (is_valid : felt):
 end
 
 @storage_var
-func authenticator() -> (authenticator_address : felt):
+func authenticators(authenticator_address : felt) -> (is_valid : felt):
+end
+
+@storage_var
+func executor() -> (executor_address : felt):
```
pscott, line 44: Executor is set per space now, but I think we could imagine having an `executor` per proposal (we'd simply need to change this to take as input a `proposal_id`)

contracts/starknet/space/space.cairo
```diff
@@ -52,56 +64,124 @@ end
...
+        metadata_uri_len : felt, metadata_uri : felt*, execution_params_len : felt,
+        execution_params : felt*):
 end
 
 @event
 func vote_created(proposal_id : felt, voter_address : EthAddress, vote : Vote):
 end
 
+# Internal utility function to hash data
+func hash_pedersen{pedersen_ptr : HashBuiltin*}(calldata_len : felt, calldata : felt*) -> (
```
Orland0x: The starware.py  and starknet.js methods for hashing an array append the length of the array to the end of itself before hashing. Not sure why they did it that way (was debugging for 3 hours trying to figure out why the hashes were not matching lol), but we should do this too so the hash can be easily computed offchain.

Orland0x:
Im wondering whether the entire proposal should be cancelled if overflow on the voting power count happens rather than just making that vote fail. The proposal is broken if it happens so seems like it should be cancelled

contracts/starknet/space/space.cairo
```diff
@@ -119,14 +197,12 @@ func vote{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr : fe
         assert 1 = 0
```
Orland0x: can use `with_attr error_message("message"):` for assertions with a specific failure error message

contracts/starknet/space/space.cairo
```diff
@@ -52,56 +62,124 @@ end
...
-        voter_address : EthAddress, proposal_id : felt, choice : felt, params_len : felt,
-        params : felt*) -> ():
+        voting_strategy_contract : felt, voter_address : EthAddress, proposal_id : felt,
+        choice : felt, voting_params_len : felt, voting_params : felt*) -> ():
     alloc_locals
 
     # Verify that the caller is the authenticator contract.
-    authenticator_only()
+    assert_valid_authenticator()
+    assert_valid_voting_strategy(voting_strategy_contract)
```
Orland0x: The way I saw it, There is a set of whitelisted voting strategies set in the space and when you vote, `get_voting_power` will be called on each strategy in the list. The overall voting power for the voter will be the sum of voting powers from each strategy. voters do not get to choose what strategy they use

test/starknet/shared/helpers.ts
```diff
@@ -17,3 +32,74 @@ export function bytesToHex(bytes: number[]): string {
   }).join('');
   return '0x' + body;
 }
+
+export async function setup() {
```
Orland0x: do you think it makes sense to put the setup function here rather than in the test file itself? I was using `helpers.ts` as a place to put helper functions used by all tests - could move a lot of it to sx.js at somepoint too

contracts/starknet/space/space.cairo
pscott: very good idea :) Had this on my todo for a long time so here it is: 2e6deb8ea8da9cd42a005c229136474c2d25c886

contracts/ethereum/SnapshotXZodiacModule/SnapshotXL1Executor.sol
pscott: @bonustrack do you need the set of whitelisted spaces emitted too?

contracts/starknet/space/space.cairo
pscott: The way I saw it, we would be giving this to `calldata` (meaning actual calldata would be `calldata + calldata_len`). This reduces computation on-chain, at the cost of adding a little bit of complexity off-chain (the hash data provided to starkware.py will not be the same one submitted to the contract). Wdyt?

pscott: The way I saw it, we would be giving this to `calldata` (meaning actual calldata would be `calldata + calldata_len`). This reduces computation on-chain, at the cost of adding a little bit of complexity off-chain (the hash data provided to starkware.py will not be the same one submitted to the contract). Wdyt?

pscott: The way I saw it, we would be giving this to `calldata` (meaning actual calldata would be `calldata + calldata_len`). This reduces computation on-chain, at the cost of adding a little bit of complexity off-chain (the hash data provided to starkware.py will not be the same one submitted to the contract). Wdyt?

pscott: The way I saw it, we would be giving this to `calldata` (meaning actual calldata would be `calldata + calldata_len`). This reduces computation on-chain, at the cost of adding a little bit of complexity off-chain (the hash data provided to starkware.py will not be the same one submitted to the contract). Wdyt?

contracts/starknet/space/space.cairo
pscott: thanks! 8a221a86c31608186f1bdd54c0c7388e88878beb

pscott: thanks! 8a221a86c31608186f1bdd54c0c7388e88878beb

pscott: thanks! 8a221a86c31608186f1bdd54c0c7388e88878beb

pscott: > The way I saw it, There is a set of whitelisted voting strategies set in the space and when you vote, `get_voting_power` will be called on each strategy in the list. The overall voting power for the voter will be the sum of voting powers from each strategy. voters do not get to choose what strategy they use

We've talked about it yesterday on Discord. For future reference, we've decided that:
1) If an overflow happens (or any other logic error happens while voting), we don't automatically cancel the proposal. It is up to the `controller` to decide what to do in this scenario.
2) We will however need to implement a `cancel_proposal` function, only callable by the `controller`, which will basically be like `finalize_proposal`. This `cancel_proposal` should call the execution contract, and we should rename `hasPassed` to something like `proposalStatus` or `proposalOutcomeStatus`, and have different status `ACCEPTED, REJECTED, CANCELLED`. This will allow execution strategies to handle the logic differently :)

pscott: > The way I saw it, There is a set of whitelisted voting strategies set in the space and when you vote, `get_voting_power` will be called on each strategy in the list. The overall voting power for the voter will be the sum of voting powers from each strategy. voters do not get to choose what strategy they use

We've talked about it yesterday on Discord. For future reference, we've decided that:
1) If an overflow happens (or any other logic error happens while voting), we don't automatically cancel the proposal. It is up to the `controller` to decide what to do in this scenario.
2) We will however need to implement a `cancel_proposal` function, only callable by the `controller`, which will basically be like `finalize_proposal`. This `cancel_proposal` should call the execution contract, and we should rename `hasPassed` to something like `proposalStatus` or `proposalOutcomeStatus`, and have different status `ACCEPTED, REJECTED, CANCELLED`. This will allow execution strategies to handle the logic differently :)

pscott: > The way I saw it, There is a set of whitelisted voting strategies set in the space and when you vote, `get_voting_power` will be called on each strategy in the list. The overall voting power for the voter will be the sum of voting powers from each strategy. voters do not get to choose what strategy they use

We've talked about it yesterday on Discord. For future reference, we've decided that:
1) If an overflow happens (or any other logic error happens while voting), we don't automatically cancel the proposal. It is up to the `controller` to decide what to do in this scenario.
2) We will however need to implement a `cancel_proposal` function, only callable by the `controller`, which will basically be like `finalize_proposal`. This `cancel_proposal` should call the execution contract, and we should rename `hasPassed` to something like `proposalStatus` or `proposalOutcomeStatus`, and have different status `ACCEPTED, REJECTED, CANCELLED`. This will allow execution strategies to handle the logic differently :)

contracts/starknet/space/space.cairo
Orland0x: oh okay I see what you mean. Yeah happy to do it this way. Maybe add a comment where its used so its super clear https://github.com/snapshot-labs/sx-core/blob/2e6deb8ea8da9cd42a005c229136474c2d25c886/contracts/starknet/space/space.cairo#L289

contracts/starknet/space/space.cairo
pscott: > Im wondering whether the entire proposal should be cancelled if overflow on the voting power count happens rather than just making that vote fail. The proposal is broken if it happens so seems like it should be cancelled

We've talked about it yesterday on Discord. For future reference, we've decided that:
1) If an overflow happens (or any other logic error happens while voting), we don't automatically cancel the proposal. It is up to the `controller` to decide what to do in this scenario.
2) We will however need to implement a `cancel_proposal` function, only callable by the `controller`, which will basically be like `finalize_proposal`. This `cancel_proposal` should call the execution contract, and we should rename `hasPassed` to something like `proposalStatus` or `proposalOutcomeStatus`, and have different status `ACCEPTED, REJECTED, CANCELLED`. This will allow execution strategies to handle the logic differently :)

contracts/ethereum/SnapshotXZodiacModule/ProposalRelayer.sol
pscott: removed in 6e560643a1cfc0d511668f7c15e049449c9f8118 :D

contracts/starknet/space/space.cairo
pscott: implemented in https://github.com/snapshot-labs/sx-core/pull/67/commits/e41d8bf32a0fef85620b8bec941c46d50d8b10fb

contracts/starknet/space/space.cairo
```diff
@@ -52,92 +67,221 @@ end
...
     next_proposal_nonce.write(1)
 
     return ()
 end
 
 @external
 func vote{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr : felt}(
-        voter_address : EthAddress, proposal_id : felt, choice : felt, params_len : felt,
-        params : felt*) -> ():
+        voting_strategy_contract : felt, voter_address : EthAddress, proposal_id : felt,
```
Orland0x: `voting_strategy_contract` shouldnt be passed here, same in `propose`

contracts/starknet/space/space.cairo
```diff
@@ -171,38 +321,165 @@ func propose{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr :
     let end_timestamp = start_timestamp + duration
 
     # Get the voting power of the proposer
-    let (strategy_contract) = voting_strategy.read()
     let (voting_power) = IVotingStrategy.get_voting_power(
```
Orland0x: will use cumulative voting power here like in vote

pscott: woups! thanks :D bc4b99aabd59a3a854c77eeab2431e0a1c3a4a82

contracts/starknet/space/space.cairo
pscott: hehe indeed, dead code :) 908226416fa83d9f210c2182f00c6e5bcbe665ac

contracts/starknet/space/space.cairo
pscott: good idea! :) 37783bc73643bb0cf17a3fcad9eef5704849bfef

test/starknet/shared/helpers.ts
pscott: Hmm I can see this `setup` function being used in a couple of different tests (and actually I believe it's already the case), that's why I put it here in the helpers.ts. I could however make a `setup.ts` rather than putting it here in the `helper.ts`. wdyt?

Orland0x: I kind of prefer it in the test file (even if it gets duplicated) as i see it as part of the test rather than just a helper function.

Orland0x: yeah nice idea having a `setup.ts` file with the various setup functions in would make more sense. Then  we have `vanillaSetup()` (this one) and others for different configs...

Orland0x: And we keep `helpers.ts` for functions that dont use hardhat - which we should be able to move to sx.js soon

contracts/ethereum/SnapshotXZodiacModule/SnapshotXL1Executor.sol
Orland0x: lets leave for now, can add in future

test/starknet/shared/helpers.ts
pscott: [b801cf1](https://github.com/snapshot-labs/sx-core/pull/67/commits/b801cf19de7c8c57b309d2260b04a571153cea13)

bonustrack (approved):

Orland0x (approved):

PR #76, opened 2022-04-07
Use arrays for authenticators and voting strategies

pscott:
Instead of using a single authenticator and a single voting strategy, spaces can now pass in arrays of authenticators and voting strategies to the space constructor.


pscott:
Merged in #67

PR #77, opened 2022-04-10
Add more details about project in README

pscott:
This PR adds details in the README, mainly by giving an overview of the flow and the different modules of Snapshot X.
It also renames `voting_period` to `voting_duration` which better represents the variable.


bonustrack:
How "duration" is better than "period"? I've used period cuz its how it's called on Compound governor, I'd prefer if we follow same naming.

README.md
```diff
@@ -20,16 +20,49 @@
...
 
-Usage
------
+#### Authenticators
+
+Authenticators are the contracts in charge of authenticating users. This repository provides two useful authenticators:
+- [ethereum_authenticator](contracts/starknet/authenticator/ethereum.cairo) which will authenticate a user based on a message signed by Ethereum private keys.
+- [starknet_authenticator](contracts/starknet/authenticator/starknet.cairo), which will authenticate a user based on a message signed by Starknet keys.
```
bonustrack: I would call it StarkKey Sig authenticator, cuz it's not exactly specific to StarkNet, it could be used for some L3 or StarkEx. And I'm adding "Sig" cuz we could have later an authenticator for StarkKey tx.
```suggestion
- [starkkey_sig_authenticator](contracts/starknet/authenticator/StarkKeySig.cairo), which will authenticate a user based on a message signed by StarkNet keys.

```

README.md
```diff
@@ -20,16 +20,49 @@
...
 
-Snapshot is open-sourced software licensed under the © [MIT license](LICENSE).
+Voting strategy contracts are the contracts used to determine the voting power of a user. The most common example is using the erc20 token balance of a user to determine his voting power. But we could imagine other voting strategies: owning a specific NFT, owning NFT of collection X and another NFT of collection Y, having participated in protocol xyz... the possibilities are endless! We provide the [single_slot_proof](contracts/starknet/strategies/single_slot_proof.cairo) which allows classic ERC20 and ERC721 balances on L1 (thanks to [fossil](https://github.com/OilerNetwork/fossil)) to be used as voting power, but feel free to create your own strategies! The interface of a strategy can be found [here](contracts/starknet/strategies/interface.cairo).
 
-Usage
------
+#### Authenticators
+
+Authenticators are the contracts in charge of authenticating users. This repository provides two useful authenticators:
+- [ethereum_authenticator](contracts/starknet/authenticator/ethereum.cairo) which will authenticate a user based on a message signed by Ethereum private keys.
```
bonustrack: I would add "Sig" in the name to differentiate with the other authenticator that use Ethereum transaction from L1. Also I'm thinking the authenticator could be used for Polygon, Avalanche etc, we may want to call it differently than Ethereum, it could be based on the algo name instead like ECDSA what do you think?

README.md
```diff
@@ -20,16 +20,125 @@
     </a>
 </div>
 
-## DISCLAIMER: STILL IN DEVELOPMENT
+## Overview
 
-This project is still under heavy development. Feel free to contact us on [Discord](https://discord.snapshot.org)!
+Governance mechanisms obey a trilemma between **decentralisation**, **cost**, and **flexibility**. On the one end you have systems like [Snapshot](https://snapshot.org/#/) that provide a very cheap experience [1] with lots of flexibility. However it relies on one trusting the Snapshot off-chain protocol to deliver the verdict of a particular governance proposal and to not censor votes. Flexibility comes from the wide range of voting strategies than one can employ to calculate the voting power for each user.
```
pscott: I believe it's "On one hand" / "On the other hand" ?

README.md
```diff
@@ -20,16 +20,125 @@
...
 
-## License
+On the other end you have governance systems that run fully on-chain on Ethereum mainnet. [Compound Governor](https://github.com/compound-finance/compound-protocol/blob/master/contracts/Governance/GovernorBravoDelegate.sol) is one example of such a system. All the voting logic is computed on-chain which provides an equivalent degree of decentralisation to Ethereum itself. The compromise of such a system is a high cost of participation due to the high gas costs incurred when transacting on the blockchain. The flexibility of the system is also limited by cost as sophisticated voting strategies require increased on-chain logic and therefore it would cost even more and limit participation further to utilise them.
```
pscott: "On the other hand"?

README.md
pscott: https://dictionary.cambridge.org/dictionary/english/on-the-one-hand-on-the-other-hand

pscott:
> How "duration" is better than "period"? I've used period cuz its how it's called on Compound governor, I'd prefer if we follow same naming.

When we talked about it on Discord the other day we all agreed that `duration` made more sense than `period`. The only reason we'd keep `period` would be to "follow the compound standard" but since we're building some different tools, I'm not sure we should stick to their naming IF we find a better naming solution :) wdyt?

README.md
```diff
@@ -20,16 +20,125 @@
...
+
+### Fossil Storage Verifier
+
+The backbone of the voting strategies is the Fossil module built by the awesome Oiler team. This module allows any part of the Ethereum mainnet state to be trustlessly verified on StarkNet. Verification of Ethereum state information is achieved by submitting a proof of the state to StarkNet and then verifying that proof. Once this state information has been proved, we can then calculate voting power as an arbitrary function of the information. For more information on Fossil, refer to their [Github](https://github.com/OilerNetwork/fossil)
```
pscott: proven* 
love this paragraph :)

README.md
Orland0x: yeah, sounds better

README.md
Orland0x: thanks :)

Orland0x:
> > How "duration" is better than "period"? I've used period cuz its how it's called on Compound governor, I'd prefer if we follow same naming.
> 
> When we talked about it on Discord the other day we all agreed that `duration` made more sense than `period`. The only reason we'd keep `period` would be to "follow the compound standard" but since we're building some different tools, I'm not sure we should stick to their naming IF we find a better naming solution :) wdyt?

yeah i agree that keeping to a convention isnt that important here, not too bothered either way though

README.md
```diff
@@ -20,16 +20,125 @@
...
+
+### Authenticators
+
+Authenticators are the contracts in charge of authenticating users. All authenticators have a function titled `execute` that takes the following arguments: 
+- `to`: The address of the space contract the user wants to interact with.
```
pscott, line 86: @bonustrack @Orland0x I would prefer to use `target` rather than `to`. I know `to` is used by other contracts (I think Argent), but I find `target` to be much more concise. Wdyt?

README.md
```diff
@@ -20,16 +20,125 @@
...
+
+### Authenticators
+
+Authenticators are the contracts in charge of authenticating users. All authenticators have a function titled `execute` that takes the following arguments: 
+- `to`: The address of the space contract the user wants to interact with. 
+- `function selector`: The function selector for the function inside the space contract that the user wants to invoke (vote or propose).
+- `calldata`: An array of parameters that are required by the function specified by `function selector`. 
+
+Beyond this, each authenticator implements different logic depending on the type of authentication that is being done. This repository provides three useful authenticators:
+- [Ethereum Signature Authenticator](contracts/starknet/authenticator/ethereum.cairo): Will authenticate a user based on a message signed by Ethereum private keys.
```
pscott, line 91: @bonustrack I'm tagging you here regarding the comment you made saying we should use something like `ECDSA` rather than Ethereum...
The problem with that is the curve used is `secp256k1` so I think it should be `ECDSA secp256k1` which is pretty long...
We could maybe go for `EVM Signature Authenticator`? wdyt?

README.md
```diff
@@ -20,16 +20,125 @@
...
+
+Upon successful authentication of the user, the `execute` method will call the function specified by `function selector` in the space contract `to`, with `calldata` as arguments. 
+
+This modularity allows spaces to authenticate users using other authentication methods: For example if you wanted to use Solana keys to authenticate users, you would simply need to write the authenticator contract on Starknet, whitelist it in your space, and Solana users would be able to vote in your DAO!
+
+### Execution
+
+The `executor` contract implements the execution strategy that gets called when voting for a proposal is done. The interface can be found [here](contracts/starknet/execution/interface.cairo). 
+Once voting has ended, calling `finalize_proposal` in the space contract will pass the following parameters to the `execute` method of the `executor`:
+- `has_passed`: Whether the proposal has passed or not (majority of voting power allocated to `FOR` signifies that the proposal has passed).
```
pscott: we've modified this to `proposal_outcome` now

README.md
```diff
@@ -20,16 +20,125 @@
...
+- `execution_hash`: Hash of the transactions to be executed.
+- `execution_params`: Array of additional parameters that are needed by the execution strategy. 
+
+Currently this repo provides the Zodiac Execution Strategy. This enables a DAO perform permissionless execution of L1 Gnosis Safe transactions upon the successful completion of a proposal. The [Zodiac Relayer](contracts/starknet/execution/zodiac_relayer.cairo`) contract is the `executor` for this strategy, which will forward the execution to the [L1 Zodiac module](contracts/ethereum/SnapshotXZodiacModule/SnapshotXL1Executor.sol) address specified in `executions_params[0]`. To use this strategy, DAOs will need a Gnosis Safe with the Snapshot X Zodiac module activated.
```
pscott: a DAO to* perform

README.md
bonustrack: @pscott Never mind we can go for Ethereum signature, I feel EVM would be more confusing

README.md
```diff
@@ -20,16 +20,125 @@
...
+
+We will now briefly provide an overview of the off-chain aspects of Snapshot X and how they will interact with the on-chain code. 
+
+<p align="center">
+<img src="./docs/images/offchain_architecture.png" width="800">
+</p>
+
+### Snapshot X UI
+
+The [UI](https://github.com/snapshot-labs/sx-ui) provides a simple interface for users to interact with Snapshot X. This will guide users through deploying spaces, creating proposals, and voting. It will also access data from the API and act as a DAO governance dashboard. In future we plan to fully integrate the UI into the existing Snapshot UI to provide a seamless experience between the various governance options offered by Snapshot Labs.
```
pscott, line 126: In the* future

README.md
bonustrack: Maybe we should follow what OZ is doing with Cairo Account Abstraction, but not strong opinion here

README.md
pscott: Yeah. Agreed Ethereum is not THE ideal solution, but I feel like it's currently the best...

README.md
Orland0x: "In future" seems fine here id say

README.md
pscott: So OZ uses `to` (see https://github.com/OpenZeppelin/cairo-contracts/blob/dfafdd2ebd778718cf440555e975f6eace278695/src/openzeppelin/account/library.cairo#L26-L29). But it's not very explicit (I believe the `Call` structure is in the library part of the code...).

I find code readability more useful than following "almost standards", and let's say I have a "moderately strong opinion" here haha :p

README.md
pscott: TIL: https://english.stackexchange.com/questions/2005/do-in-future-and-in-the-future-imply-different-meanings#:~:text=In%20future%20is%20commonly%20used,ring%20up%20beforehand%20in%20future.

README.md
Orland0x: shall we change the cairo file to `ethereum_signature.cairo` then? And while we are at it, probably better to call L1 transaction authenticator `Ethereum Transaction Authenticator` and cairo file `ethereum_transaction.cairo`

Orland0x: I sort of use L1 and Ethereum synonomously which might not be to clear to some people

pscott: Yeah I think we should try to use `Ethereum` as much as possible, and keep `l1` really only for technical stuff / places where it's useful to have "short" naming?

README.md
Orland0x: "Can you send 1 eth to the to address" 👎 
"Can you send 1 eth to the target address" 👍  

Latter is much nicer for sure

README.md
Orland0x: agreed

Orland0x: same with Starknet and L2

README.md
pscott: I'm imagining the "drake meme" on this one hahah

Orland0x (approved):

bonustrack (approved):

PR #79, opened 2022-04-11
Interact via l1

Orland0x:
Implemented the necessary contracts to allow interactions with SX via an L1 tx. 
- `StarkNetCommit.sol` : Allows one to commit a hash of a tx to Starknet.
- `eth_tx.cairo` : Authenticates an interaction with SX by checking that the hash of the tx matches the hash committed from L1.

Also: 
- Added library function to convert a felt to 4 words. I had used this in a initial implementation of `l1_tx.cairo`, but didnt actually need it. I left it in the repo as it should come in handy at some point. 
- Fixed small error in `single_slot_proof.cairo` that Scott picked up last PR but I forgot to change.


contracts/ethereum/L1Interact/StarkNetCommit.sol
```diff
@@ -0,0 +1,60 @@
...
+ * @dev This contract is designed to be a generic standard that that can be used by any StarkNet protocol that wants to allow interactions via an L1 transaction.
+ */
+contract StarkNetCommit is Initializable {
+  /// The StarkNet core contract.
+  // IStarknetCore public immutable starknetCore;
+  /// Using a mock here for testing purposes
+  MockStarknetMessaging public immutable starknetCore;
+
+  /// Address of the L1 tx authenticator contract
+  uint256 public authL1;
```
pscott: I feel like `authL1` doesn't convey enough information.
I thought about `starknetAuth` but it's the same problem. I think we need to be explicit here:
`starknetAuthenticatorOfL1Tx` or something like that. Sure it's very verbose, but at least we know it's a starknet contract that verifies L1 tx... wdyt?

contracts/ethereum/L1Interact/StarkNetCommit.sol
```diff
@@ -0,0 +1,60 @@
...
+    authL1 = _authL1;
+  }
+
+  /**
+   * @dev Commit a hash and the sender address to StarkNet.
+   * @param _hash The hash to commit
+   */
+  function commit(uint256 _hash) external {
+    uint256[] memory payload = new uint256[](2);
+    payload[0] = uint256(uint160(msg.sender));
```
pscott, line 56: out of curiosity, why do we need to use `uint160` here? Solidity doesn't supoprt `uin256t(msg.sender)` ?

test/l1l2/l1_tx.ts
```diff
@@ -0,0 +1,228 @@
...
+
+const propose_selector = BigInt(
+  '0x1BFD596AE442867EF71CA523061610682AF8B00FC2738329422F4AD8D220B81'
+);
+const vote_selector = BigInt('0x132BDF85FC8AA10AC3C22F02317F8F53D4B4F52235ED1EABB3A4CBBE08B5C41');
+const VOTING_DELAY = BigInt(0);
+const VOTING_PERIOD = BigInt(20);
+const RANDOM_ADDRESS = BigInt('0xAD4Eb63b9a2F1A4D241c92e2bBa78eEFc56ab990');
+
+export async function setup(signer: SignerWithAddress) {
```
pscott: We will be able to move this function to the `setup.ts` file once the other PR gets merged :D

test/l1l2/l1_tx.ts
```diff
@@ -0,0 +1,228 @@
...
+    // const vote_commit = getCommit(target, vote_selector, vote_calldata);
+    // await starknetCommit.commit(vote_commit);
+    // await starknet.devnet.flush();
+
+    // // Casting vote
+    // await l1TxAuthenticator.invoke('execute', {
+    //   target: target,
+    //   function_selector: vote_selector,
+    //   calldata: vote_calldata,
+    // });
```
pscott: ?

contracts/ethereum/L1Interact/StarkNetCommit.sol
Orland0x: yep cant cast directly: https://stackoverflow.com/questions/43318077/solidity-type-address-not-convertible-to-type-uint256

test/l1l2/l1_tx.ts
Orland0x: not needed, will remove

contracts/ethereum/L1Interact/StarkNetCommit.sol
Orland0x: yeah see what you mean `authL1` isnt clear. yeah happy to go with the super explicit name

.github/workflows/tests.yml
```diff
@@ -44,15 +44,18 @@ jobs:
         run: |
           sudo apt install -y libgmp3-dev
           python -m pip install --upgrade pip
-          pip install starknet-devnet==0.1.23
+          pip install starknet-devnet==0.1.19
```
pscott, line 47: any particular reason as to why we're downgrading starknet-devnet ?

.github/workflows/tests.yml
```diff
@@ -44,15 +44,18 @@ jobs:
...
-          curl http://localhost:8000 &&
-          yarn test 
+      - name: Run Cross Chain Tests
+        run: yarn test:crosschain:ci
+          
+      - name: Run L1 tests
+        run: yarn test:l1 
+        
+      - name: Run L2 tests 
+        run: yarn test:l2:ci
```
pscott, line 61: Good idea!

scripts/test-crosschain.sh
```diff
@@ -0,0 +1,18 @@
...
+yarn wait-on tcp:8000 && 
+wait-on tcp:8545 && 
+yarn hardhat test test/crosschain/zodiac.ts --network 'ethereumLocal' --starknet-network 'starknetLocal' && 
+sleep 10 && 
+yarn hardhat test test/crosschain/ethereum_transaction_authentication.ts --network 'ethereumLocal' --starknet-network 'starknetLocal'
+if [ $? -eq 0 ]
+then
+  kill -9 $(lsof -t -i:8545)
+  kill -9 $(lsof -t -i:8000)
+  echo here
```
pscott: Maybe change to `echo "Error"` or something more explicit?

contracts/ethereum/L1Interact/StarkNetCommit.sol
```diff
@@ -0,0 +1,60 @@
+/// SPDX-License-Identifier: MIT
+
+pragma solidity 0.8.9;
+
+import '@openzeppelin/contracts/proxy/utils/Initializable.sol';
+// import './Interfaces/IStarknetCore.sol';
```
pscott, line 6: Path is incorrect; Should we maybe add the `IstarknetCore.sol` to the ethereum contracts, and provide an accurate path here? So that the day we wish to activate it we just need to comment / uncomment?

contracts/starknet/authenticator/eth_tx.cairo
```diff
@@ -0,0 +1,78 @@
...
+    let (input_array : felt*) = alloc()
+    assert input_array[0] = target
+    assert input_array[1] = function_selector
+    memcpy(input_array + 2, calldata, calldata_len)
+    # Appending the length of the array to itself as the offchain version of the hash works this way
+    assert input_array[calldata_len + 2] = calldata_len + 2
+    let (hash_state_ptr) = hash_init()
+    let (hash_state_ptr) = hash_update{hash_ptr=pedersen_ptr}(
+        hash_state_ptr, input_array, calldata_len + 3
+    )
```
pscott: I think we could share this code: https://github.com/snapshot-labs/sx-core/blob/0f81eddad82008762e4e6faf0205f337fdb0481f/contracts/starknet/space/space.cairo#L107-L118 by moving it to `lib/hash_pedersen` and call it here? wdyt?

pscott (approved):

contracts/starknet/authenticator/eth_tx.cairo
Orland0x: yes would make sense. although, i think we should add the length to the array inside the contract rather than  offchain so that the interface of `execute` is accurate. As including the length, `calldata` is actually more like `calldata_plus_length`. The step savings of doing it offchain is pretty negligible.

pscott: 👍 go for it!

contracts/ethereum/L1Interact/StarkNetCommit.sol
Orland0x: yeah will do

.github/workflows/tests.yml
Orland0x: it doesnt work with anything more recent in the GA so its sort of rough fix for now. super annoying

contracts/ethereum/L1Interact/StarkNetCommit.sol
Orland0x: `'./Interfaces/IStarknetCore.sol';` is actually correct. theres different interfaces for starknet core in each folder. 

could just have one but the zodiac module and starknet commit use different functions in StrarknetCore so i just kept them separate.

pscott: Oh yeah right I had a look on the current `develop` branch, didn't see you added it with the PR lol. Nice :D

Issue #99, opened 2022-04-29
Use of EthAddress in interfaces


pscott:
Currently, some interfaces / functions are using `EthAddress` because the MVP was made with L1 transactions in mind.
However in the future, we  might want to change to `Uint256` (and not felt, because some keys on other chains might not fit in a `felt`!).

(This issue refers to the `get_voting_power` interface for example, or the `vote` and `propose` function of the space contract :))

Orland0x:
Using a general `Address` type

PR #100, opened 2022-04-29
Add whitelist strategy

### LINKED ISSUE #92: Add whitelist strategy

pscott:
- Adds a simple whitelist strategy. The smart contract has no owner: whitelisted addresses are declared in the constructor. If we wish to add / remove addresses, just create a new smart contract (as this is a voting strategy and won't hold any funds, I think it's safe to to so).
- As mentioned in #99 , we will need to transition to something else than `EthAddress` as the type for the `get_voting_power` interface because we might also have starkkeys, or solana keys, or bitcoin keys etc...

Closes #92

### DIFF
--- a/contracts/starknet/voting_strategies/whitelist.cairo
+++ b/contracts/starknet/voting_strategies/whitelist.cairo
@@ -19,12 +19,12 @@ func register_whitelist{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_
     else:
         let address = EthAddress(_whitelist[0])
         # Add it to the whitelist
-        whitelist.write(address, 1)
+        whitelist.write(address, _whitelist[1])
 
         # Emit event
         whitelisted.emit(address)
 
-        register_whitelist(_whitelist_len - 1, &_whitelist[1])
+        register_whitelist(_whitelist_len - 2, &_whitelist[2])
         return ()
     end
 end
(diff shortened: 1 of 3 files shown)


contracts/starknet/strategies/whitelist.cairo
```diff
@@ -0,0 +1,49 @@
...
+
+@storage_var
+func whitelist(address : EthAddress) -> (is_valid : felt):
+end
+
+@event
+func whitelisted(address : EthAddress):
+end
+
+func register_whitelist{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr : felt}(
```
bonustrack: One easy (or not) improvement would be to add a parameter here to define the weight, this way you can give more power for certain addresses, make it much more flexible

pscott: fd830c419fa4f543a0c68356b4f8daedec680bd4 :D

contracts/starknet/voting_strategies/whitelist.cairo
```diff
@@ -0,0 +1,48 @@
...
+    if _whitelist_len == 0:
+        return ()
+    else:
+        let address = EthAddress(_whitelist[0])
+        # Add it to the whitelist
+        let vp = Uint256(_whitelist[1], _whitelist[2])
+        whitelist.write(address, vp)
+
+        # Emit event
+        whitelisted.emit(address)
```
Orland0x: shall we emit the voting power of the address here too?

Orland0x (approved):

contracts/starknet/voting_strategies/whitelist.cairo
pscott: good idea! 7a3f8d871276878a1f1b32e907b9d5b3d71cd9ed :D

PR #102, opened 2022-05-01
Restructure repo

Orland0x:
Repo Restructuring and Renaming to increase consistency and readability.


.github/workflows/lint.yml
```diff
@@ -46,4 +46,4 @@ jobs:
       
       - name: Run Linters
         run: |
-          yarn check-format
+          yarn format
```
pscott: I think we should stick with `check-format` here and have the CI only throw an error when code is not formatted properly. If we run `format` here, basically it will format the code for the CI, but not commit it, so our codebase will remain unformatted I think

contracts/starknet/interfaces/i_authenticator.cairo
```diff
@@ -0,0 +1,7 @@
+%lang starknet
+
+@contract_interface
+namespace i_authenticator:
```
pscott, line 4: inb4 @bonustrack fires you for writing this 👺👺👺

pscott (approved):

.github/workflows/lint.yml
Orland0x: oh i see yes will revert

contracts/starknet/interfaces/i_authenticator.cairo
Orland0x: i just love snake case so much, cant help it...

PR #103, opened 2022-05-03
Whitelist executors

### LINKED ISSUE #94: Move `execution` from a per space to a per proposal

pscott:
This PR:
- Moves the `executor` from a single address to a whitelist of executors (provided via constructor, can be modified by the `controller` using the `add_executors` and `remove_executors`)
- Proposals now are required to have an `executor` specified (when creating a proposal via `propose`). This `executor` needs to be part of the whitelist
- When calling `finalize_proposal`, the executor is checked to see whether it's still in the whitelist. If it is not, then the `proposal_outcome` is set to `CANCELLED`

Closes #94


pscott:
Side note:
I've created 3 issues that I thought about when writing this code:
- Add the ability to add / remove voting strategies and authenticators (code is missing lol, but till be very similar to `add_executors` and `remove_executors`) (https://github.com/snapshot-labs/sx-core/issues/105)
- Use the `Owner` from OZ contracts rather than our own custom implementation (https://github.com/snapshot-labs/sx-core/issues/104)

contracts/starknet/space.cairo
```diff
@@ -92,9 +92,9 @@ func only_controller{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_che
 
     let (_controller) = controller.read()
 
-    with_attr error_message("You are not the controller"):
-        assert caller_address = _controller
-    end
+    # with_attr error_message("You are not the controller"):
+    #     assert caller_address = _controller
+    # end
```
pscott: I get this weird error where `caller_address` != `_controller` when running the tests. I think it used to work but not sure. When I print the address, `_controller` has indeed been set correctly, so it looks like the problem comes from `get_caller_address` but it seems weird. Need to investigate some more...

pscott: had to call `.starknetContract.address` that's why :)

contracts/starknet/space.cairo
```diff
@@ -113,6 +113,38 @@ func update_controller{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_c
     return ()
 end
 
+@external
+func add_executors{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr : felt}(
```
Orland0x, line 117: can we get away with just `add_executors` or just `register_executors` ? seems kind of redundant to have both

pscott: lmao I'm stupid! Ofc we can :) 843ad89e8578e0277571fd40cbcb0d15554a4c63

contracts/starknet/space.cairo
```diff
@@ -338,6 +371,12 @@ func propose{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr :
     # Verify that the caller is the authenticator contract.
     assert_valid_authenticator()
 
+    # Verify that the executor address is one of the whitelisted addresses
```
Orland0x, line 394: could be cleaner to have an `assert_valid_executor()` function like we have for authentication

contracts/starknet/space.cairo
```diff
@@ -443,10 +482,17 @@ func finalize_proposal{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_c
...
     end
 
-    let (executor_address) = executor.read()
+    let (is_valid) = executors.read(proposal.executor)
+    if is_valid == 0:
+        # Executor has been removed from the whitelist. Cancel this execution.
+        tempvar proposal_outcome = ProposalOutcome.CANCELLED
+    else:
+        # Classic cairo reference hackz
+        tempvar proposal_outcome = proposal_outcome
```
Orland0x, line 508: nice xd

contracts/starknet/space.cairo
pscott: you are right! 703e0dabfee218371ed6b2e8f48d96f629498c7f

contracts/starknet/space.cairo
pscott: Oh actually the reason why there was an `add_executors` and a `register_executors` was because `add_executors` is `only_controller()` whereas `register` is not. Since `_controller` is an argument passed in to the `constructor` (and not `get_caller_address`), `add_executors` fails when called in the `constructor`... I think we should keep `register_executors`, but we could decide to go the other way and simply set the `_controller` to `get_caller_address` in the constructor

Orland0x: ok i see yes. annoying but not much we can do then

Orland0x: once we create the space factory we can properly figure out the space deployment/controller update flows

pscott: Decided to fix it by having `add_executors` simply be a wrapper with a call to `only_owner()` before `register_executors`. I think naming should be changed though, and maybe use `_` as we sometimes do in solidity :)

23440683e044789d60d27dfa6f5071436ebe4c97

Orland0x (approved):

PR #110, opened 2022-05-05
Use OZ ownable functions

### LINKED ISSUE #104: Use OZ `Ownable` standard rather than `only_controller`
I think we should use [OZ's standard](https://github.com/OpenZeppelin/cairo-contracts/blob/main/src/openzeppelin/access/ownable.cairo) rather than our own standard for managing access :)

pscott:
Closes #104

### DIFF
--- a/contracts/starknet/space.cairo
+++ b/contracts/starknet/space.cairo
@@ -19,6 +19,10 @@ from contracts.starknet.lib.choice import Choice
 from contracts.starknet.lib.proposal_outcome import ProposalOutcome
 from contracts.starknet.lib.hash_array import hash_array
 
+from openzeppelin.access.ownable import (
+    Ownable_only_owner, Ownable_transfer_ownership, Ownable_get_owner
+)
+
 @storage_var
 func voting_delay() -> (delay : felt):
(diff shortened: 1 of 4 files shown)


contracts/starknet/space.cairo
```diff
@@ -84,31 +88,20 @@ func vote_created(proposal_id : felt, voter_address : EthAddress, vote : Vote):
...
-    with_attr error_message("You are not the controller"):
-        assert caller_address = _controller
-    end
-
-    return ()
+func controller_updated(new_controller : felt, previous_controller : felt):
 end
 
+@external
 func update_controller{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr : felt}(
```
Orland0x, line 91: should we set the controller in the constructor?

pscott: Yeah I need to remove the `controller` storage_var all together, haven't done that I'm still fighting with a bug on https://github.com/snapshot-labs/sx-core/pull/103 hahaha

Orland0x (approved):

Issue #91, opened 2022-04-28
Emit an event when a space is created


pscott:

pscott:
@bonustrack @Orland0x I don't remember why I created this issue: is it a request from you guys or just something I thought about and forgot to properly document ? 🙈

bonustrack:
This is a request from me, if we issue an event with space settings when the space is deployed we can catch this info with Checkpoint to populate these settings on the API. Currently it's hardcoded.

pscott:
@bonustrack got it :) easy to add! :)

Orland0x:
cool, the event will be emitted in the space factory once that exists though right?

bonustrack:
Yep

PR #111, opened 2022-05-08
Add min/max duration and quorum

### LINKED ISSUE #95: Add minVotingPeriod and maxVotingPeriod parameters

pscott:
Adds min and max voting duration:
- min_voting_duration: a proposal can't be finalized before `min_votin_duration` has elapsed
- max_voting_duration: a vote can't be cast after `max_voting_duration` has elapsed

I'd like to add more tests but currently we can't manipulate the block timestamp in starknet-devnet. We might want to add some later on though :) 

Closes #95


contracts/starknet/lib/proposal.cairo
```diff
@@ -4,7 +4,6 @@ from starkware.cairo.common.uint256 import Uint256
 struct Proposal:
     member execution_hash : Uint256
     member start_timestamp : felt
-    member end_timestamp : felt
```
pscott, line 7: I've removed that and just compute the end timestamp now, saving us one write :D :)

bonustrack: But then if someone change period the proposal will also change period?

contracts/starknet/space.cairo
```diff
@@ -278,17 +283,18 @@ func constructor{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_p
...
+        assert_le(_min_voting_duration, _max_voting_duration)
         assert_not_zero(_executors_len)
     end
     # TODO: maybe use uint256_signed_nn to check proposal_threshold?
 
     # Initialize the storage variables
     voting_delay.write(_voting_delay)
-    voting_duration.write(_voting_duration)
+    min_voting_duration.write(_min_voting_duration)
+    max_voting_duration.write(_max_voting_duration)
```
bonustrack: Would be nice to add event here with all these params if its easy to do or can be done in another PR

contracts/starknet/lib/proposal.cairo
pscott: Hmm you're right. Currently there's no way of editing those so I thought it wasn't an issue but maybe I should already write the code with this in mind, so that when the day comes and we decide to make those editable, we don't introduce any bugs :)

contracts/starknet/space.cairo
pscott: was thinking about doing this in another PR (to close https://github.com/snapshot-labs/sx-core/issues/91) :)

contracts/starknet/lib/proposal.cairo
pscott: 339b15a11f2691287728e2d798c374af44a711e8

bonustrack: Great

Orland0x:
I think we should implement quorum before adding this feature

pscott:
> I think we should implement quorum before adding this feature

0503bfb5bc5935469bcbdd3a49806c512ad2033e :) (missing tests, they will come in tomorrow hopefully :D)

pscott:
Latest commits add the `quorum` storage var. It is stored on the space, and also stored on each new `proposal` so that if `quorum` changes during proposal, it doesn't disturb any of the ongoing proposals.

contracts/starknet/space.cairo
```diff
@@ -94,7 +103,8 @@ end
 @event
 func space_created(
```
Orland0x, line 104: quorum emitted in the event?

Orland0x (approved):

contracts/starknet/space.cairo
pscott: 0fbf8d42ae5f935eef06387eead974266418fa9c :D

pscott:
Also added a function to edit quorum as discussed on Discord (23e86a73314f1dff86a8d1e8daf6aae0822ed605) :D

PR #117, opened 2022-05-09
Add starknet auth

pscott:
This PR aims at adding a starknet auth (and testing it)

Current tasks:
- [x] Add starknet auth
- [ ] Testing starknet auth

The message checked is `hash_pedersen(calldata)`. This is fine for the `vote` (since it has a `proposal_id` and one can't double vote). However this is not fine for `proposal` since it doesn't hold a `proposal_id`. Need to think about that :)

### DIFF
--- a/contracts/starknet/authenticators/eth_tx.cairo
+++ b/contracts/starknet/authenticators/eth_tx.cairo
@@ -43,7 +43,14 @@ end
 
 @external
 func execute{syscall_ptr : felt*, range_check_ptr, pedersen_ptr : HashBuiltin*}(
-    target : felt, function_selector : felt, calldata_len : felt, calldata : felt*
+    target : felt,
+    function_selector : felt,
+    calldata_len : felt,
+    calldata : felt*,
+    signer_len : felt,
+    signer : felt*,
+    signature_len : felt,
+    signature : felt*,
 ):
     alloc_locals
     # Cast arguments to single array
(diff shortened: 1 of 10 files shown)


contracts/starknet/authenticators/eth_tx.cairo
```diff
@@ -43,7 +43,14 @@ end
...
 @external
 func execute{syscall_ptr : felt*, range_check_ptr, pedersen_ptr : HashBuiltin*}(
-    target : felt, function_selector : felt, calldata_len : felt, calldata : felt*
+    target : felt,
+    function_selector : felt,
+    calldata_len : felt,
+    calldata : felt*,
+    signer_len : felt,
+    signer : felt*,
+    signature_len : felt,
```
bonustrack, line 52: Any reason to have signature for eth tx?

pscott: The interface has been modified so every authenticator should have this now (can be set to 0 if the auth doesn't need it)

bonustrack: Why would you want same interface for authenticators?

pscott: I don't know, there currently is an `i_authenticator.cairo` so I thought we had good reasons to want an interface. I guess if we wish to make sure our contracts are all easily integratable then it makes sense to have an authenticator interface, but it's true we can do without it.

bonustrack: I imagine these authenticators more external to SX, others apps may want to use these auths I'd prefer each auth to be custom tailored for their use case and not depending on the parameters of others auths

pscott:
Closed in favor of https://github.com/snapshot-labs/sx-core/pull/138

PR #121, opened 2022-05-11
Slot Key Verification

Orland0x:
Added slot key verification to the single slot strategy so that users can only submit proofs for the correct slots. 

To do this I had to add another parameter called `global_voting_strategy_params` which is an array stored for each voting strategy set in the space. `global_voting_strategy_params`  has broader use case: it is fully general purpose and can be used to customize any voting strategy defined with specific parameters. 

For single slot proof, `global_voting_stategy_params` is structured as follows: 
- `global_voting_stategy_params[0]` = contract address where storage is read from. (eg ERC20 token contract address)
- `global_voting_stategy_params[1]` = slot index of the mapping where storage is being read from. 

Also added a new data structure called `Immutable2DArray` which is used to hold `global_voting_strategy_params` and `voting_strategy_params` for all strategies (array of arrays type structure)


contracts/starknet/interfaces/i_voting_strategy.cairo
```diff
@@ -6,7 +6,12 @@ from starkware.cairo.common.uint256 import Uint256
 @contract_interface
 namespace i_voting_strategy:
     func get_voting_power(
-        timestamp : felt, address : EthAddress, params_len : felt, params : felt*
+        block : felt,
```
pscott, line 9: may I ask why it got changed to a `block`?

contracts/starknet/lib/array2d.cairo
```diff
@@ -0,0 +1,35 @@
...
+func get_sub_array{range_check_ptr}(array2d : Immutable2DArray, index : felt) -> (
+    array_len : felt, array : felt*
+):
+    let offset = array2d.offsets[index]
+    let array = &array2d.elements[offset]
+    if index == array2d.offsets_len - 1:
+        tempvar array_len = array2d.elements_len - offset
+    else:
+        tempvar array_len = array2d.offsets[index + 1] - offset
+    end
```
pscott: Maybe I'm sm0l brain but I think adding a little comment here to explain what we're doing would be beneficial for people reading the code! :) (It's understandable, but adding some comment will put more clarity as to what exactly we're doing :))

contracts/starknet/lib/felt_to_uint256.cairo
```diff
@@ -0,0 +1,23 @@
...
+
+# Convert a felt to 2 128 bit words returned as a uint256.
+func felt_to_uint256{bitwise_ptr : BitwiseBuiltin*, range_check_ptr}(value : felt) -> (
+    uint256 : Uint256
+):
+    let (low) = bitwise_and(value, MASK_LOW)
+
+    let (t1) = bitwise_and(value, MASK_HIGH)
+    let (t1, _) = unsigned_div_rem(t1, SHIFT_120)
+    let (high, _) = unsigned_div_rem(t1, SHIFT_8)
```
pscott, line 20: I fail to see why we don't directly bit shift by `128`, but maybe I'm completely missing something?

contracts/starknet/lib/slot_key.cairo
```diff
@@ -0,0 +1,23 @@
+from starkware.cairo.common.cairo_builtins import BitwiseBuiltin
+from starkware.cairo.common.alloc import alloc
+from starkware.cairo.common.keccak import unsafe_keccak
+from starkware.cairo.common.uint256 import Uint256
+from starknet.lib.felt_to_uint256 import felt_to_uint256
+
+# Returns the EVM slot key for mappings (key can be any value type)
+# UNSAFE: This is not safe to use in production code due to unsafe keccak - waiting for safe version
+func get_slot_key{bitwise_ptr : BitwiseBuiltin*, range_check_ptr}(
```
pscott, line 11: It would be nice to provide maybe some context as to `slot_index` and `mapping_key` refer to? Or give a link?

contracts/starknet/lib/slot_key.cairo
```diff
@@ -0,0 +1,23 @@
...
+) -> (slot_key : Uint256):
+    alloc_locals
+    let (encoded_array : felt*) = alloc()
+    let (slot_index_uint256) = felt_to_uint256(slot_index)
+    let (mapping_key_uint256) = felt_to_uint256(mapping_key)
+    encoded_array[0] = mapping_key_uint256.high
+    encoded_array[1] = mapping_key_uint256.low
+    encoded_array[2] = slot_index_uint256.high
+    encoded_array[3] = slot_index_uint256.low
+    let (low, high) = unsafe_keccak(encoded_array, 64)
```
pscott: prefer `16 * 4` as it make it more evident that it's because we have 4 elements in encoded_array

contracts/starknet/lib/words64_to_uint256.cairo
```diff
@@ -0,0 +1,13 @@
+from starkware.cairo.common.uint256 import Uint256
+
+const SHIFT_64 = 2 ** 64
+
+# Convert a 4 words of 64 bits per felt to a Uint256 (128 bits per felt)
```
pscott: Convert 4 words of 64 bits* (remove `a`)

contracts/starknet/space.cairo
```diff
@@ -261,14 +279,54 @@ end
...
+    else:
+        # Store global voting parameter
+        global_voting_strategy_params.write(
+            voting_strategy, index, _global_voting_strategy_params[0]
+        )
+
+        if _global_voting_strategy_params_len == 1:
+            # Nothing left to add, end recursion
+            return ()
+        else:
```
pscott: You can remove that, it's actually less step efficient :)

pscott: When I say that I mean l:317 - l:320

contracts/starknet/space.cairo
```diff
@@ -387,6 +462,27 @@ func get_cumulative_voting_power{syscall_ptr : felt*, pedersen_ptr : HashBuiltin
...
+}(_voting_strategy_contract : felt, _global_voting_strategy_params : felt*, index : felt) -> (
+    global_voting_strategy_params_len : felt
+):
+    let (global_voting_strategy_param) = global_voting_strategy_params.read(
+        _voting_strategy_contract, index
+    )
+    if global_voting_strategy_param == 0:
+        return (index)
+    else:
+        assert _global_voting_strategy_params[index] = global_voting_strategy_param
```
pscott: I think we should put a comment here saying that we're actually assigning rather than `assert`-ing (even though it's the same thing in cairo). Or maybe remove the `assert`. This is because I believe the semantics around `x = y` is not the same as `assert x = y` and tbh I had to read this a couple of times in order to understand what was going on haha

contracts/starknet/space.cairo
```diff
@@ -387,6 +462,27 @@ func get_cumulative_voting_power{syscall_ptr : felt*, pedersen_ptr : HashBuiltin
...
+
+        let (global_voting_strategy_params_len) = get_global_voting_strategy_params(
+            _voting_strategy_contract, _global_voting_strategy_params, index + 1
+        )
+        return (0)
```
pscott: Maybe I'm missing something but returning `0` seems odd to me? This means this function will always return `0`?

contracts/starknet/space.cairo
```diff
@@ -355,27 +413,44 @@ func get_cumulative_voting_power{syscall_ptr : felt*, pedersen_ptr : HashBuiltin
...
 
+    # Initialize empty array to store global voting params
+    let (global_voting_strategy_params : felt*) = alloc()
+
+    # Retrieve global voting strategy params
+    let (global_voting_strategy_params_len) = get_global_voting_strategy_params(
+        voting_strategy, global_voting_strategy_params, 0
+    )
+
+    # Extract voting params array for the voting strategy specified by the index
```
pscott, line 417: Does using the `index` here work? 
while `get_cumulative_voting_power` used to iterate over all the voting strategies, now it only iterates through the `used_voting_strategies`. I *think* we need to change `index`, and actually find a way to find the corresponding `index` that corresponds to the `voting strategy`, right?

contracts/starknet/space.cairo
```diff
@@ -261,14 +279,54 @@ end
...
+
+        # Add global voting params
+        register_global_voting_strategy_params(0, to_add[0], global_params_len, global_params)
+
+        unchecked_add_voting_strategies(to_add_len - 1, &to_add[1], global_params_all, index + 1)
+        return ()
+    end
+end
+
+func register_global_voting_strategy_params{
```
pscott: How about `unchecked_add_global_voting_strategy_params` ? To follow the other `unchecked_add` convention?

contracts/starknet/voting_strategies/single_slot_proof.cairo
```diff
@@ -5,7 +5,8 @@ from starkware.cairo.common.uint256 import Uint256, uint256_add
 from starkware.cairo.common.math import unsigned_div_rem, assert_nn_le
 from contracts.starknet.fossil.contracts.starknet.types import StorageSlot
 from contracts.starknet.lib.eth_address import EthAddress
-
+from contracts.starknet.lib.slot_key import get_slot_key
+from contracts.starknet.lib.words64_to_uint256 import words64_to_uint256
 # FactRegistry simplified interface
```
pscott, line 12: Small detail but would appreciate a newline between the imports and the comment :)

contracts/starknet/voting_strategies/single_slot_proof.cairo
```diff
@@ -39,21 +40,37 @@ end
 @view
 func get_voting_power{
```
pscott, line 43: Maybe the expected `global_voting_params` should be defined here in the comments? I think it needs to be documented somewhere (for every new votingn strategy) so I think here is a good place :D

test/starknet/lib_array2d.ts
```diff
@@ -0,0 +1,63 @@
...
+    './contracts/starknet/test_contracts/test_array2d.cairo'
+  );
+  const testArray2d = await testArray2dFactory.deploy();
+  return {
+    testArray2d: testArray2d as StarknetContract,
+  };
+}
+
+describe('2D Arrays:', () => {
+  it('The library should be able to contruct the 2D array type from a flat array and then retrieve the sub arrays individually.', async () => {
```
pscott: construct*

test/starknet/lib_array2d.ts
```diff
@@ -0,0 +1,63 @@
...
+    const arr3 = [BigInt(1), BigInt(2), BigInt(3)];
+    const arr4 = [BigInt(7), BigInt(9)];
+    const arr2d = [arr1, arr2, arr3, arr4];
+    const flatArray = flatten2DArray(arr2d);
+
+    const { array: array1 } = await testArray2d.call('test_array2d', {
+      flat_array: flatArray,
+      index: 0,
+    });
+    expect(array1).to.deep.equal([BigInt(5)]);
```
pscott: I think we should `.to.deep.equal(arr1)`, in case we ever decide to change `arr1` :)

contracts/starknet/voting_strategies/single_slot_proof.cairo
```diff
@@ -39,21 +40,37 @@ end
...
-):
+}(
+    block : felt,
+    voter_address : EthAddress,
+    global_params_len : felt,
+    global_params : felt*,
+    params_len : felt,
+    params : felt*,
+) -> (voting_power : Uint256):
     alloc_locals
```
pscott, line 53: Should we first start by checking that `global_params_len == 2` ? I think currently the tests are running with this set to 0 (see my previous comment above), and still passing, because we're not checking this value?

pscott (changes requested):
Solid! Just a couple of nitpicks here and there.
One thing I would like to change though is naming! I think we need to come up with better naming, as I had a tough time differentiating `global_params` vs `global_params_all` etc etc.

To make sure I have a good understanding, what we have now is:
- `voting_strategy_params`: user-defined array of data
- `global_params`: voting strategy specific array, recorded in the `global_voting_strategy_params` storage var
- `global_params_all` and `voting_strategy_params_all`: a 2d immutable array consisting of all the different `global_params`. (`voting_strategy_params_all` is used in `get_cumulative_voting_power` and `global_params_all` in `register_global_voting_strategy_params`)

Is that correct?

Orland0x:
> Solid! Just a couple of nitpicks here and there. One thing I would like to change though is naming! I think we need to come up with better naming, as I had a tough time differentiating `global_params` vs `global_params_all` etc etc.
> 
> To make sure I have a good understanding, what we have now is:
> 
> * `voting_strategy_params`: user-defined array of data
> * `global_params`: voting strategy specific array, recorded in the `global_voting_strategy_params` storage var
> * `global_params_all` and `voting_strategy_params_all`: a 2d immutable array consisting of all the different `global_params`. (`voting_strategy_params_all` is used in `get_cumulative_voting_power` and `global_params_all` in `register_global_voting_strategy_params`)
> 
> Is that correct?

yeah this is what i used. i used `params` and `global_params` when in the frame of reference of the voting strategy (as voting strategy is implied) and `voting_strategy_params` and `global_voting_strategy_params` params when in the space contract frame.

contracts/starknet/lib/felt_to_uint256.cairo
Orland0x: theres some limit of the divisor around 2^120 im pretty sure. bit annoying

contracts/starknet/lib/array2d.cairo
Orland0x: yep will do

contracts/starknet/interfaces/i_voting_strategy.cairo
Orland0x: fossil uses block numbers and we currently dont have a way to convert so we provide a block number here. Its on my to do list to add the conversion though, then will revert.

contracts/starknet/lib/slot_key.cairo
Orland0x: yep ill reference my repo https://github.com/snapshot-labs/evm-slot-key-verification

contracts/starknet/lib/slot_key.cairo
Orland0x: yep agree

contracts/starknet/space.cairo
Orland0x: oh nice

contracts/starknet/space.cairo
Orland0x: i thought `x = y` was equivalent to `assert x = y`? i kind of use them interchangably but yeah would make more sense to stick to  `x = y` for assignments

contracts/starknet/space.cairo
Orland0x: the function should never get there so it'll never return that, just put it there as you need a return statement. maybe this is a dumb way to do it?

contracts/starknet/space.cairo
Orland0x: but now `voting_strategy_params_all` only contains the params for the strategies specified in `used_voting_strategies` so it should be ok?

Orland0x: maybe we should rename `voting_strategy_params_all` to `used_voting_strategy_params` in this context

contracts/starknet/space.cairo
Orland0x: yep agree

contracts/starknet/voting_strategies/single_slot_proof.cairo
Orland0x: yep agree

test/starknet/lib_array2d.ts
Orland0x: yep my bad

contracts/starknet/voting_strategies/single_slot_proof.cairo
Orland0x: yep will add

contracts/starknet/space.cairo
pscott: Haha yes, they are the same in cairo but what I was tryingn to say with `semantics` is that, for a developer reading the code, he/she necessarily interpret both as being equal. I thought here we wanted to `assert` (meaning make sure that `_global_voting_strategy_params[index]` was equal to `global_voting_strategy_param` but what we're doing is indeed assigning a value to that... so I think removing the `assert` will make it more clear that we're doing just that! :)

contracts/starknet/space.cairo
Orland0x: i did this ^

contracts/starknet/lib/felt_to_uint256.cairo
pscott: Ah! Might be worth adding a comment to explain this :p

contracts/starknet/voting_strategies/single_slot_proof.cairo
```diff
@@ -49,13 +53,17 @@ func get_voting_power{
...
     let (local fact_registry_addr) = fact_registry_store.read()
 
     # Decoding voting strategy parameters
     let (slot, proof_sizes_bytes_len, proof_sizes_bytes, proof_sizes_words_len, proof_sizes_words,
         proofs_concat_len, proofs_concat) = decode_param_array(params_len, params)
 
-    # Extracting global parameters for this strategy
+    # Checking that the global parameters array is valid and then extracting the individual parameters
+    # For the single slot proof strategy, the global parameters array is length 2 where the first element is the
+    # contract address where the desired slot resides, and the section element is the index of the slot in that contract.
```
pscott, line 63: section element? You mean `second`?

contracts/starknet/voting_strategies/single_slot_proof.cairo
```diff
@@ -49,13 +53,17 @@ func get_voting_power{
...
+    assert global_params_len = 2
```
pscott: Maybe add
```
with_attr error_message("Incorrect length for global parameters"):
```
?

contracts/starknet/voting_strategies/single_slot_proof.cairo
Orland0x: yep lol

contracts/starknet/voting_strategies/single_slot_proof.cairo
Orland0x: yep

contracts/starknet/space.cairo
pscott: I guess I don't understand something here. This function is supposed to return the `length` of the `_global_voting_strategy_params`, right?

In the case where `global_voting_strategy_param == 0` (meaning we've reached the end of the array) it returns `index` which seems fine to me

However I don't get why it should never return go in the `else` branch (and hence `return (0)`)?

pscott: My intuition is that this function should actually return `global_voting_strategy_params_len` but maybe I'm wrong ^^

pscott:
> > Solid! Just a couple of nitpicks here and there. One thing I would like to change though is naming! I think we need to come up with better naming, as I had a tough time differentiating `global_params` vs `global_params_all` etc etc.
> > To make sure I have a good understanding, what we have now is:
> > 
> > * `voting_strategy_params`: user-defined array of data
> > * `global_params`: voting strategy specific array, recorded in the `global_voting_strategy_params` storage var
> > * `global_params_all` and `voting_strategy_params_all`: a 2d immutable array consisting of all the different `global_params`. (`voting_strategy_params_all` is used in `get_cumulative_voting_power` and `global_params_all` in `register_global_voting_strategy_params`)
> > 
> > Is that correct?
> 
> yeah this is what i used. i used `params` and `global_params` when in the frame of reference of the voting strategy (as voting strategy is implied) and `voting_strategy_params` and `global_voting_strategy_params` params when in the space contract frame.

What would you think about remove `strategy` (to make it less verbose) and use:
- `user_voting_params`: user-defined array of data
- `space_voting_params`: (or could be `protocol_voting_params` or `strategy_voting_params` or `constant_voting_params`. I think `global` sounds like it's common for all voting strategies?)

And maybe keeping the same naming for space / voting_strategies context, as it makes it easier to navigate the code and not get lost. Not sure about this last one, just my thoughts :)

Orland0x:
> > > Solid! Just a couple of nitpicks here and there. One thing I would like to change though is naming! I think we need to come up with better naming, as I had a tough time differentiating `global_params` vs `global_params_all` etc etc.
> > > To make sure I have a good understanding, what we have now is:
> > > 
> > > * `voting_strategy_params`: user-defined array of data
> > > * `global_params`: voting strategy specific array, recorded in the `global_voting_strategy_params` storage var
> > > * `global_params_all` and `voting_strategy_params_all`: a 2d immutable array consisting of all the different `global_params`. (`voting_strategy_params_all` is used in `get_cumulative_voting_power` and `global_params_all` in `register_global_voting_strategy_params`)
> > > 
> > > Is that correct?
> > 
> > 
> > yeah this is what i used. i used `params` and `global_params` when in the frame of reference of the voting strategy (as voting strategy is implied) and `voting_strategy_params` and `global_voting_strategy_params` params when in the space contract frame.
> 
> What would you think about remove `strategy` (to make it less verbose) and use:
> 
> * `user_voting_params`: user-defined array of data
> * `space_voting_params`: (or could be `protocol_voting_params` or `strategy_voting_params` or `constant_voting_params`. I think `global` sounds like it's common for all voting strategies?)
> 
> And maybe keeping the same naming for space / voting_strategies context, as it makes it easier to navigate the code and not get lost. Not sure about this last one, just my thoughts :)

I think its important we have `strategy` in the name as otherwise its not clear that they refer to parameters for voting strategies. Snapshot X is a voting protocol after all, so `voting_params` could conceivably refer to many different things. 

yeah `user_voting_params` makes sense I agree. 

hmmmm i see your point. Although to me, `space_voting_params` sounds more like it refers to something common to all strategies than `global_voting_params` does. Ill try to think of a better name...

yep i can keep same names with the space/voting strategy context. 

I think all of these ambiguities will be solved if we have a good natspec standard frankly

Orland0x:
we could do `user_voting_strategy_params` and then just `voting_strategy_params` for the global ones?

pscott:
> we could do `user_voting_strategy_params` and then just `voting_strategy_params` for the global ones?

yes! There are three different variables though: the `user defined`, the `storage_var` (which takes all `voting_strategy_parameters` and what the `voting_strategy_parameters` storage_var holds.
Maybe we could do 

- `user_voting_strategy_params` 
- `voting_strategies_params` (for the storage var?)
- and the `voting_strategies_params` would store some `voting_strategy_params` ? 


The only issue I see is that it's hard to differentiate... or we could also do
- `user_voting_strategy_params`
- `global_voting_strategy_params` for the storage var
- and the storage var would store some `voting_strategy_params`

? wdyt?

Orland0x:
> > we could do `user_voting_strategy_params` and then just `voting_strategy_params` for the global ones?
> 
> yes! There are three different variables though: the `user defined`, the `storage_var` (which takes all `voting_strategy_parameters` and what the `voting_strategy_parameters` storage_var holds. Maybe we could do
> 
> * `user_voting_strategy_params`
> * `voting_strategies_params` (for the storage var?)
> * and the `voting_strategies_params` would store some `voting_strategy_params` ?
> 
> The only issue I see is that it's hard to differentiate... or we could also do
> 
> * `user_voting_strategy_params`
> * `global_voting_strategy_params` for the storage var
> * and the storage var would store some `voting_strategy_params`
> 
> ? wdyt?

why not just `voting_strategy_params_store`  for the storage var? tbh we could just make the `store` suffix a standard for every storage var in the contract. seems to be what starkware do anyway

pscott:
> why not just `voting_strategy_params_store` for the storage var? tbh we could just make the `store` suffix a standard for every storage var in the contract. seems to be what starkware do anyway

I haven't seen this "standard" but i think it would be an excellent idea indeed. Should it be `store` or `storage`? (`store` sounds like it could be interpreted as a verb / action?)

pscott:
I know I'm being a little "picky" here, but I do think that we need to be careful about those as we will probably be setting the standard for the rest of the codebase and the future, and naming is an important and difficult task :p 

After all, as Phil Karlton said, *There are only two hard things in Computer Science: cache invalidation and naming things.* !

Orland0x:
> > why not just `voting_strategy_params_store` for the storage var? tbh we could just make the `store` suffix a standard for every storage var in the contract. seems to be what starkware do anyway
> 
> I haven't seen this "standard" but i think it would be an excellent idea indeed. Should it be `store` or `storage`? (`store` sounds like it could be interpreted as a verb / action?)

 cool lets do that then, `store` is generally used as a noun (like 'im going to store to get some eggs') so i think it makes sense slightly more sense to use here

Orland0x:
> I know I'm being a little "picky" here, but I do think that we need to be careful about those as we will probably be setting the standard for the rest of the codebase and the future, and naming is an important and difficult task :p
> 
> After all, as Phil Karlton said, _There are only two hard things in Computer Science: cache invalidation and naming things._ !

completely agree, crucial we get this space contract right.

Orland0x:
*think* all the fixes are in

pscott (approved):
💪

contracts/starknet/space.cairo
```diff
@@ -279,58 +279,49 @@ end
...
-            &_global_voting_strategy_params[1],
-        )
+        unchecked_add_voting_strategy_params(to_add, params_len - 1, &params[1], index + 1)
         return ()
     end
 end
 
+# NOTE: We need to think carefully about how to handle the case where a voting strategy is removed.
+# Do we also need to remove the voting strategy params? Currently we are not.
+# Setting the voting strategy to 0 will remove it from the store, but what happens if someone re-adds the strategy but with different params?
```
pscott, line 317: Btw I think we should indeed remove the associated voting strategy params :)

PR #138, opened 2022-05-19
Add starknet auth

pscott:
WIP


pscott:
Will be kind of blocked by https://github.com/OpenZeppelin/cairo-contracts/issues/328 :( 

(it's not a hard block, we can get away by passing the selector and not using the interface :))

bonustrack:
Did we considered to not use any authenticator for StarkNet tx? And use space contract directly, since we don't really need to authenticate anything with a StarkNet tx

pscott:
> Did we considered to not use any authenticator for StarkNet tx? And use space contract directly, since we don't really need to authenticate anything with a StarkNet tx

We still to verify that the caller is the same as `proposer_address` / `voter_address`, so we indeed need to go through an authenticator I believe :(

Orland0x (approved):

Orland0x (approved):

PR #148, opened 2022-05-25
Add starknet execution

### LINKED ISSUE #93: Add l2 execution
Almost done, need to debug :p

pscott:
This PR adds what we call "starknet execution": the ability to create a proposal and specify some starknet transactions to automatically execute if this proposal passes.
The execution parameter layout is a bit tricky but is completely detailed in the `starknet.cairo` execution strategy file.

Here are the desired steps in order to get this right:

- [x] Basic execution (create a proposal that will in turn create a new proposal if executed. Execute it and verify that the new proposal has been created)
- [x] Verify the hash
- [ ] Add a test to modify internal settings (like quorum)
- [ ] Add a test to execute MULTIPLE transactions
- [ ] Decide what to do with the results of the successive `calls` (imho nothing, but still we need to ask ourselves the question)

Closes #93

### DIFF
--- a/contracts/starknet/execution_strategies/starknet.cairo
+++ b/contracts/starknet/execution_strategies/starknet.cairo
@@ -39,7 +39,8 @@ func execute_calls{syscall_ptr : felt*}(data_ptr : felt*, calls_len : felt, call
 
     # TODO: what should we do with the return values?
 
-    # do the next calls recursively
+    # Do the next calls recursively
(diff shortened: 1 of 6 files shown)


pscott:
Linting is not working because I've updated the cairo compiler (and formatter) version to `0.8.2` locally on my machine but the work is on another PR (https://github.com/snapshot-labs/sx-core/pull/138).

PR #159, opened 2022-05-30
few updates

Orland0x:
Few updates. Got a bit carried away, will do on separate PRs next time 

Allow handling of both Ethereum and Starknet addresses in the space contract and voting strategies. 

Switched to a felt to uint256 function from cairo common

Fixed vulnerability in authenticators where the library function `execute` could be called directly. 

Added starknet.js deployment script for a space contract with a number of modules activated. Can call deployment script with `yarn deploy:goerli`. 
Deployment info in `./deployments/goerli1.json`


contracts/starknet/lib/general_address.cairo
```diff
@@ -0,0 +1,4 @@
+# Generalized type used to represent addresses in Snapshot X. Eg Ethereum, Starknet, etc.
+struct Address:
+    member value : felt
+end
```
pscott, line 4: Should this be called `lib/address` ? I find it odd to have a filename be `general_address` and the actual object be `Address` ?

test/crosschain/EthTxAuth.test.ts
```diff
@@ -138,14 +132,12 @@ describe('L1 interaction with Snapshot X', function () {
     await starknet.devnet.loadL1MessagingContract(networkUrl, mockStarknetMessaging.address);
     await starknetCommit
       .connect(signer)
-      .commit(
-        getCommit(BigInt(space.address), BigInt(getSelectorFromName('vote')), proposeCalldata)
-      ); // Wrong selector
+      .commit(getCommit(BigInt(space.address), PROPOSE_SELECTOR, proposeCalldata)); // Wrong selector
```
pscott, line 135: This was supposed to be `VOTE_SELECTOR`, not `PROPOSE_SELECTOR` I think (since the comment says "Wrong selector") ?

test/starknet/StarkTxAuth.test.ts
```diff
@@ -1,75 +1,100 @@
...
+  let proposeCalldata: bigint[];
+
+  // Additional parameters for voting
+  let voterAccount: Account;
+  let voterAddress: string;
+  let proposalId: bigint;
+  let choice: Choice;
+  let usedVotingStrategies2: bigint[];
+  let userVotingParamsAll2: bigint[][];
+  let voteCalldata: bigint[];
```
pscott, line 41: nice and tidy

pscott (approved):
All good except one test that got modified :)

test/crosschain/EthTxAuth.test.ts
Orland0x: The test was testing that revert case (ie authentication fails if the wrong hash is committed) - the comment is just referring to what I changed to cause it to fail.

contracts/starknet/lib/general_address.cairo
Orland0x: hmm yeah its not ideal, just thought that people would assume its a Starknet address if its called `address.cairo`

PR #183, opened 2022-06-14
Timestamp to block number

Orland0x:
With this update, the snapshot of a proposal is made using a timestamp rather than a block number. Specifically, the snapshot is given by the block timestamp at proposal creation obtained via `get_block_timestamp()`. This allows SX to be natively multi-chain as timestamps are universal whereas block numbers are chain dependent. 

Certain voting strategies do however require a block number, such as the single slot proof strategy. This resolution between timestamp and block number should occur at the *voting strategy level* and is up to the strategy creator to ensure the validity of the conversion. 

For the single slot proof strategy, we provide this functionality within `lib/timestamp.cairo`. It works as follows: 
- When a proposal is created, the snapshot timestamp is passed to the single slot voting strategy. 
- Fossil is then queried for the latest ethereum block number stored within it, and this number is returned to the voting strategy and stored in a mapping `timestamp_to_eth_block_number`.
- When a user casts a vote, again the same snapshot timestamp is passed to the single slot voting strategy.
-  The `timestamp_to_eth_block_number`mapping is then queried with this timestamp which will return the same ethereum block number as before. We therefore ensure that the snapshot block number of the proposal remains constant even if later block hashes are sent to fossil during the period of the proposal.

We also add more extensive testing of the single slot proof strategy in this PR.


contracts/starknet/VotingStrategies/SingleSlotProof.cairo
```diff
@@ -70,7 +75,11 @@ func get_voting_power{
     # contract address where the desired slot resides, and the section element is the index of the slot in that contract.
     assert params_len = 2
     let contract_address = params[0]
-    let slot_index = params[1]
+
+    # In the current implementation, we cant store arrays with a zero as the last element because the read function would just treat
+    # the array is 1 element shorter. Hence, we offset the slot_index by 1 so that it is never zero.
+    # Probably worth changing our array handling so this is not necessary.
+    let slot_index = params[1] - 1
```
Orland0x, line 82: this isnt ideal. we could change how we store arrays.

contracts/starknet/Space.cairo
```diff
@@ -759,26 +761,22 @@ func propose{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr :
...
     # Verify that the caller is the authenticator contract.
     assert_valid_authenticator()
 
     # Verify that the executor address is one of the whitelisted addresses
     assert_valid_executor(executor)
 
-    let (current_timestamp) = get_block_timestamp()
+    # The snapshot for the proposal is the current timestamp at proposal creation
+    # We use a timestamp instead of a block number to define a snapshot so that the system can generalize to multi-chain
+    let (snapshot_timestamp) = get_block_timestamp()
```
pscott, line 773: Are we absolutely 100% sure `get_block_timestamp` will return something valid (e.g not `0`)? I think yes but I'm just wondering if starknet/ware have some sort of disclaimer somewhere?

contracts/starknet/VotingStrategies/SingleSlotProof.cairo
```diff
@@ -52,7 +54,10 @@ func get_voting_power{
 ) -> (voting_power : Uint256):
     alloc_locals
 
-    let (local fact_registry_addr) = fact_registry_store.read()
+    let (fact_registry_addr) = fact_registry_store.read()
+
+    let (eth_block_number) = get_eth_block_number(timestamp)
+    let eth_block_number = eth_block_number - 1  # temp shift - waiting for Fossil fix
```
pscott, line 60: interesting, off-by-one error huh?

Orland0x, line 60: https://github.com/OilerNetwork/fossil/issues/24

contracts/starknet/lib/timestamp.cairo
```diff
@@ -0,0 +1,37 @@
...
+) -> (number : felt):
+    let (number) = timestamp_to_eth_block_number.read(timestamp)
+    if number != 0:
+        # The timestamp has already be queried in fossil and stored. Therefore we can just return the stored value
+        # This branch will be taken whenever a vote is cast as the mapping value would be set at proposal creation.
+        return (number)
+    else:
+        # The timestamp has not yet been queried in fossil. Therefore we must query Fossil for the latest eth block
+        # number stored there and store it here in the mapping indexed by the timestamp provided.
+        # This branch will be taken whenever a proposal is created, except for the (rare) case of multiple proposals
```
pscott, line 30: rare for now but could be quite common in the near future 🤝

contracts/starknet/lib/timestamp.cairo
```diff
@@ -0,0 +1,37 @@
...
+func get_eth_block_number{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(
+    timestamp : felt
+) -> (number : felt):
+    let (number) = timestamp_to_eth_block_number.read(timestamp)
+    if number != 0:
+        # The timestamp has already be queried in fossil and stored. Therefore we can just return the stored value
+        # This branch will be taken whenever a vote is cast as the mapping value would be set at proposal creation.
+        return (number)
+    else:
+        # The timestamp has not yet been queried in fossil. Therefore we must query Fossil for the latest eth block
```
pscott, line 28: This means a proposer can't create a proposal with a snapshot that was BEFORE the creation of the proposal right? Like if I create a proposal on second of Jan and I want the snapshot to be taken on the 1st of Jan (one day before), that's not possible, correct?

test/shared/parseRPCData.ts
```diff
@@ -34,13 +41,21 @@ export interface ProofInputs {
...
   accountProofSizesBytes: bigint[];
   accountProofSizesWords: bigint[];
   accountProof: bigint[];
-  userVotingPowerParams: bigint[];
+  storageProofs: bigint[][]; // Multiple storage proofs
 }
 
+/**
+ * Takes a proofs object obtained via a getStorageProof RPC call parses the data to extract the necessary data
+ * and convert it to the correct form required by the SX/Fossil contracts.
```
pscott: converts*

test/shared/setup.ts
```diff
@@ -16,7 +22,7 @@ export interface Fossil {
 }
 
 export async function vanillaSetup() {
-  const controller = (await starknet.deployAccount('OpenZeppelin')) as Account;
```
pscott, line 19: why the change from OZ to Argent? :)

pscott (approved):
Yup, good work man :) Quite some code to review but it's well written :)

contracts/starknet/VotingStrategies/SingleSlotProof.cairo
pscott: Yeah didn't comment on that since I saw this comment. Def agree 🤝

contracts/starknet/VotingStrategies/SingleSlotProof.cairo
Orland0x: ye. https://github.com/OilerNetwork/fossil/issues/24

test/shared/setup.ts
Orland0x: Made more sense I thought as we would be using argent wallets to interact with SX

contracts/starknet/lib/timestamp.cairo
Orland0x: Well technically you could if the latest block number stored in Fossil was on the 1st. This shouldnt happen though as I expect block hashes to be sent hourly. 

The ethereum block number snapshot will always be sometime before the proposal creation, the length of which depends on the regularity of block hashes being sent from L1

contracts/starknet/Space.cairo
Orland0x: Yeah I wonder whether it is enforced that the sequencer must include it. Could bring it up with them. Added a comment.

contracts/starknet/VotingStrategies/SingleSlotProof.cairo
Orland0x: opened an issue: https://github.com/snapshot-labs/sx-core/issues/188

Issue #151, opened 2022-05-26
Remove execution_hash


pscott:
We should not store the execution_hash, but simply add it in the `execution_params` array

PR #194, opened 2022-06-17
Add starknet execution

Orland0x:
Continuation of Scotts work on https://github.com/snapshot-labs/sx-core/pull/148. 

We generalised the execution for SX by removing the execution hash (that was a uint256 stored for each proposal). Now there is just the `execution_params` array that can arbitrarily contain execution hashes or simply raw tx data. For each proposal, we store the pedersen hash of this array. It is up to the execution strategy to decode this data and perform the necessary execution. 

New PR as there was lots of structural changes to the repo since the first one so decided it was easier to start again and copy code across.

### DIFF
--- a/contracts/starknet/ExecutionStrategies/Starknet.cairo
+++ b/contracts/starknet/ExecutionStrategies/Starknet.cairo
@@ -1,19 +1,17 @@
 %lang starknet
 
-from starkware.cairo.common.uint256 import Uint256, uint256_eq
 from starkware.cairo.common.math_cmp import is_le
 from starkware.starknet.common.syscalls import call_contract
 from starkware.cairo.common.alloc import alloc
-from starkware.starknet.common.messages import send_message_to_l1
 from contracts.starknet.lib.proposal_outcome import ProposalOutcome
-from contracts.starknet.lib.hash_array import hash_array
-from starkware.cairo.common.cairo_builtins import HashBuiltin
-from openzeppelin.account.library import AccountCallArray, Call
 
-# For example: say the contract wants to interact with the contract `0x123` with selector `0x456` that takes two arguments (we will give 13 and 37 as values),
+# Starknet Execution
+# This is more of a demo as we need to interact with a wallet to execute meaningful transactions.
+
+# Example execution: say the contract wants to interact with the contract `0x123` with selector `0x456` that takes two arguments (we will give 13 and 37 as values),
 # and then interact with contract `0x789` with selector `0xabc` that takes three arguments (we will give 42 43 44 as values):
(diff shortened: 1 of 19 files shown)


contracts/starknet/ExecutionStrategies/Starknet.cairo
```diff
@@ -0,0 +1,76 @@
...
+        function_selector=calls[1],
+        calldata_size=calls[2],
+        calldata=&data_ptr[calls[3]],
+    )
+
+    # TODO: what should we do with the return values?
+
+    # Do the next calls recursively
+    # We subtract `4` + calls[2]` because a `call` has 4 felts and we also need to substract its associated calldata length.
+    execute_calls(data_ptr, calls_len - (4 + calls[2]), calls + 4)
```
Orland0x, line 47: Previously it was `execute_calls(data_ptr, calls_len - (4 + calls[2]), calls)`
I think we needed to shift the pointer to the start of the calls array along by 4 here

pscott:
Closes https://github.com/snapshot-labs/sx-core/issues/151 right?

Orland0x:
> Closes #151 right?

yep

contracts/starknet/ExecutionStrategies/Starknet.cairo
```diff
@@ -0,0 +1,58 @@
...
+
+    return ()
+end
+
+@external
+func execute{syscall_ptr : felt*, range_check_ptr : felt}(
+    proposal_outcome : felt, execution_params_len : felt, execution_params : felt*
+):
+    alloc_locals
+    # Check that there are `Calls` ton execute in the execution parameters.
```
pscott, line 40: typo `ton`

scripts/deployTestSpace.ts
```diff
@@ -67,7 +75,7 @@ async function main() {
   const singleSlotProofVotingStrategyAddress = responses[3].address!;
   const vanillaExecutionStrategyAddress = responses[4].address!;
   const zodiacRelayerExecutionStrategyAddress = responses[5].address!;
-
+  const starknetExecutionStrategyAddress = responses[6].address!;
```
pscott, line 78: what does the `!` do after address?

pscott (approved):
Nice work 🤝

Orland0x:
> Nice work 🤝

team effort on this one 🤝

scripts/deployTestSpace.ts
Orland0x: its the non-null assertion operator, which tells the interpreter that this cannot be `null` or `undefined`. refused to run otherwise

pscott: galaxy brain!
TIL: [non-null operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html)

PR #198, opened 2022-06-21
Improved array storage handling in Space contract

### LINKED ISSUE #188: Handle Arrays with elements equal to zero
This will mean we do not have to offset values equal to zero when storing them.

Orland0x:
In this PR we modify how voting parameter arrays are stored in the space contract. Now, the first parameter of the array is the length of the array and then the following parameters are the elements.

This now enables us to remove the voting parameter array associated with a voting strategy when a voting strategy is removed, which we add in this PR.

This closes https://github.com/snapshot-labs/sx-core/issues/188

### DIFF
--- a/test/shared/setup.ts
+++ b/test/shared/setup.ts
@@ -368,7 +368,7 @@ export async function singleSlotProofSetup(block: any, proofs: any) {
   const minVotingDuration = BigInt(0);
   const maxVotingDuration = BigInt(2000);
   const votingStrategies: bigint[] = [BigInt(singleSlotProofStrategy.address)];
-  const votingStrategyParams: bigint[][] = [[proofInputs.ethAddressFelt, BigInt(1)]];
+  const votingStrategyParams: bigint[][] = [[proofInputs.ethAddressFelt, BigInt(0)]]; // For the aave erc20 contract, the balances mapping has a storage index of 0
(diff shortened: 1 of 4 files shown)


test/shared/setup.ts
```diff
@@ -368,7 +368,7 @@ export async function singleSlotProofSetup(block: any, proofs: any) {
   const minVotingDuration = BigInt(0);
   const maxVotingDuration = BigInt(2000);
   const votingStrategies: bigint[] = [BigInt(singleSlotProofStrategy.address)];
-  const votingStrategyParams: bigint[][] = [[proofInputs.ethAddressFelt, BigInt(1)]];
+  const votingStrategyParams: bigint[][] = [[proofInputs.ethAddressFelt, BigInt(0)]]; // For the aave erc20 contract, the balances mapping has a storage index of 0
```
pscott, line 371: Mind expanding on that? It is common? Is it expected?

pscott (approved):
🔥 as always

test/shared/setup.ts
Orland0x: for the test we are proving aave balances, and in their specific erc20 implementation the slot index of the `balances[]` mapping is 0. Not always o though, which is annoying as each DAO using it will have to work out their own slot index. Ive got a rough tool to do it in the evm-slot-key-verification repo

PR #208, opened 2022-06-28
Controller actions tests

### LINKED ISSUE #199: Full coverage on controller actions test
Add tests for the set/reset of each parameter in the space contract by the space controller.

Orland0x:
Closes #199


test/starknet/ControllerActions.test.ts
```diff
@@ -48,4 +49,54 @@ describe('Controller Actions', () => {
...
   }).timeout(600000);
+
+  it('The controller can add and remove authenticators', async () => {
+    const authenticators: bigint[] = [BigInt(1234), BigInt(4567)];
+    await controller.invoke(space, 'add_authenticators', {
+      to_add: authenticators,
+    });
+    await controller.invoke(space, 'remove_authenticators', {
+      to_remove: authenticators,
+    });
```
pscott, line 60: We're checking if adding / removing doesn't throw, but we're not checking if we can then call the space with a different authenticator. Meaning if the function `add_authenticators` was empty, we wouldn't know it from this test :p

test/starknet/ControllerActions.test.ts
```diff
@@ -48,4 +49,54 @@ describe('Controller Actions', () => {
...
+    await controller.invoke(space, 'update_voting_delay', {
+      new_delay: BigInt(1234),
+    });
+  }).timeout(600000);
+
+  it('The controller can update the min voting duration', async () => {
+    await controller.invoke(space, 'update_min_voting_duration', {
+      new_min_duration: BigInt(1234),
+    });
+  }).timeout(600000);
```
pscott, line 89: same

test/starknet/ControllerActions.test.ts
```diff
@@ -48,4 +49,54 @@ describe('Controller Actions', () => {
...
+    await controller.invoke(space, 'update_quorum', {
+      new_quorum: SplitUint256.fromUint(BigInt(1234)),
+    });
+  }).timeout(600000);
+
+  it('The controller can update the voting delay', async () => {
+    await controller.invoke(space, 'update_voting_delay', {
+      new_delay: BigInt(1234),
+    });
+  }).timeout(600000);
```
pscott, line 83: same

pscott (changes requested):
Two things are bugging me:
1) My comment regarding the fact we're not actually testing it, but rather making sure the functions we call don't throw errors. I believe it'd be better if we actually tested the BEHAVIOUR (for example when editing quorum, fetching the quorum value and making sure it's been updated. For authenticators, casting a vote with the authenticator we just added etc).
2) We're checking that the controller can "control" the SC, but I believe we should check that other addresses (meaning `!= controller`) CANNOT change. This would be useful for regression tests to make sure we don't break anything :)

pscott:
Closes #199

PR #234, opened 2022-07-18
Space as account

### LINKED ISSUE #218: Use account contract for space

Orland0x:
Implements the functionality of the space contract as an extension of an OZ account contract. The modified account is called `SpaceAccount.cairo` and the library that implements the voting logic is inside `voting_library.cairo`. We follow the OZ extensibility pattern specified here: https://github.com/OpenZeppelin/cairo-contracts/blob/main/docs/Extensibility.md 

closes https://github.com/snapshot-labs/sx-core/issues/218


contracts/starknet/lib/voting_library.cairo
```diff
@@ -0,0 +1,1066 @@
+# SPDX-License-Identifier: MIT
```
pscott: I would remove the suffix `library` because we are already in `/lib`. I think `lib/voting.cairo` would make more sense?

contracts/starknet/lib/voting_library.cairo
```diff
@@ -0,0 +1,1066 @@
...
+from contracts.starknet.lib.hash_array import hash_array
+from contracts.starknet.lib.array2d import Immutable2DArray, construct_array2d, get_sub_array
+from contracts.starknet.lib.slot_key import get_slot_key
+
+#
+# Storage
+#
+
+@storage_var
+func Voting_voting_delay_store() -> (delay : felt):
```
pscott: Can you give a little bit of context around the new naming convention?
I like the `store` suffix, but wondering why we added the `Voting` prefix. Why the prefix, and why uppercase?

contracts/starknet/lib/voting_library.cairo
```diff
@@ -0,0 +1,1066 @@
...
+        _quorum : Uint256,
+        _voting_strategy_params_flat_len : felt,
+        _voting_strategy_params_flat : felt*,
+        _voting_strategies_len : felt,
+        _voting_strategies : felt*,
+        _authenticators_len : felt,
+        _authenticators : felt*,
+        _executors_len : felt,
+        _executors : felt*,
+    ):
```
pscott: Since we're using the `store` suffix everywhere in the storage vars, I believe we could remove the `_` prefix in those arguments

contracts/starknet/lib/voting_library.cairo
```diff
@@ -0,0 +1,1066 @@
...
+        # Sanity checks
+        with_attr error_message("Invalid constructor parameters"):
+            assert_nn(_voting_delay)
+            assert_le(_min_voting_duration, _max_voting_duration)
+            assert_not_zero(_controller)
+            assert_not_zero(_voting_strategies_len)
+            assert_not_zero(_authenticators_len)
+            assert_not_zero(_executors_len)
+        end
+        # TODO: maybe use uint256_signed_nn to check proposal_threshold?
```
pscott: We need to create an issue to track this badboy

contracts/starknet/lib/voting_library.cairo
```diff
@@ -0,0 +1,1066 @@
...
+    func update_min_voting_duration{
+        syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr : felt
+    }(new_min_duration : felt):
+        Ownable.assert_only_owner()
+
+        let (previous_duration) = Voting_min_voting_duration_store.read()
+
+        let (max_duration) = Voting_max_voting_duration_store.read()
+
+        assert_le(new_min_duration, max_duration)
```
pscott: Should we add an error message here (`with_attr error_message(...)`)?

contracts/starknet/lib/voting_library.cairo
```diff
@@ -0,0 +1,1066 @@
...
+    func update_max_voting_duration{
+        syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr : felt
+    }(new_max_duration : felt):
+        Ownable.assert_only_owner()
+
+        let (previous_duration) = Voting_max_voting_duration_store.read()
+
+        let (min_duration) = Voting_min_voting_duration_store.read()
+
+        assert_le(min_duration, new_max_duration)
```
pscott: Should we add an error message here too?

contracts/starknet/lib/voting_library.cairo
```diff
@@ -0,0 +1,1066 @@
...
+                tempvar range_check_ptr = range_check_ptr
+                tempvar ecdsa_ptr = ecdsa_ptr
+                tempvar bitwise_ptr = bitwise_ptr
+            else:
+                tempvar syscall_ptr = syscall_ptr
+                tempvar pedersen_ptr = pedersen_ptr
+                tempvar range_check_ptr = range_check_ptr
+                tempvar ecdsa_ptr = ecdsa_ptr
+                tempvar bitwise_ptr = bitwise_ptr
+            end
```
pscott: can't wait till cairo gets better at this :D

contracts/starknet/lib/voting_library.cairo
```diff
@@ -0,0 +1,1066 @@
...
+            tempvar range_check_ptr = range_check_ptr
+            tempvar ecdsa_ptr = ecdsa_ptr
+            tempvar bitwise_ptr = bitwise_ptr
+        end
+
+        # Flag this proposal as executed
+        # This should not create re-entrency vulnerability because the message
+        # executor is a whitelisted address. If we set this flag BEFORE the call
+        # to the executor, we could have a malicious attacker sending some random
+        # invalid execution_params and cancel out the vote.
```
pscott: We can remove this comment I think :)

contracts/starknet/lib/voting_library.cairo
```diff
@@ -0,0 +1,1066 @@
...
+            proposal_outcome=proposal_outcome,
+            execution_params_len=execution_params_len,
+            execution_params=execution_params,
+        )
+
+        # Flag this proposal as executed
+        # This should not create re-entrency vulnerability because the message
+        # executor is a whitelisted address. If we set this flag BEFORE the call
+        # to the executor, we could have a malicious attacker sending some random
+        # invalid execution_params and cancel out the vote.
```
pscott: Can remove this one I think

contracts/starknet/lib/voting_library.cairo
```diff
@@ -0,0 +1,1066 @@
...
+        end
+
+        let proposal_outcome = ProposalOutcome.CANCELLED
+
+        IExecutionStrategy.execute(
+            contract_address=proposal.executor,
+            proposal_outcome=proposal_outcome,
+            execution_params_len=execution_params_len,
+            execution_params=execution_params,
+        )
```
pscott: If `proposal.executor == 1` (meaning it's the space account), this won't work. I think we need an `if` here to handle that case :)

contracts/starknet/lib/voting_library.cairo
```diff
@@ -0,0 +1,1066 @@
...
+
+    with_attr error_message("Invalid executor"):
+        assert is_valid = 1
+    end
+
+    return ()
+end
+
+# Computes the cumulated voting power of a user by iterating over the voting strategies of `used_voting_strategies`.
+# TODO: In the future we will need to transition to an array of `voter_address` because they might be different for different voting strategies.
```
pscott: Maybe create an issue to track this one?

README.md
```diff
@@ -35,13 +35,14 @@ contracts
 │   │  ├─ proposal_info.cairo — "Proposal vote data type"
 │   │  ├─ proposal_outcome.cairo — "The set of proposal outcomes"
 │   │  ├─ slot_key.cairo — "Function to find the slot key for a slot in the Ethereum state"
+│   │  ├─ voting.cairo — "Core library that implements the logic for voting in Snapshot X"
```
pscott, line 38: name of file here is `voting` but actual file name is `voting_library`. Imho `voting` > `voting_library` :)

contracts/starknet/SpaceAccount.cairo
```diff
@@ -0,0 +1,124 @@
...
+
+from contracts.starknet.lib.voting_library import Voting
+
+#
+# Constructor
+#
+
+@constructor
+func constructor{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(
+    public_key : felt,
```
pscott, line 19: Do we need a `public_key` for the `SpaceAccount`? As far as I understand it, the `public_key` is only used if we wish to call [is_valid_signature](https://github.com/OpenZeppelin/cairo-contracts/blob/ad399728e6fcd5956a4ed347fb5e8ee731d37ec4/src/openzeppelin/account/library.cairo#L122) or for [is_valid_eth_signature](https://github.com/OpenZeppelin/cairo-contracts/blob/ad399728e6fcd5956a4ed347fb5e8ee731d37ec4/src/openzeppelin/account/library.cairo#L149) but I don't think we'll be using those will we?

pscott: If we can manage without `public_key`, then I'm all for removing it (makes code lighter, and removes some unexpected misconfigurations where DAO creators add a `public_key` and get their corresponding private key hacked, resulting in the hack of the space contract funds :p

test/starknet/Space.test.ts
```diff
@@ -5,6 +5,12 @@ import { utils } from '@snapshot-labs/sx';
 import { vanillaSetup } from '../shared/setup';
 import { PROPOSE_SELECTOR, VOTE_SELECTOR } from '../shared/constants';
 
+import { hash } from 'starknet';
+const { getSelectorFromName } = hash;
+
+const P = BigInt(getSelectorFromName('__execute__'));
+console.log(P);
```
pscott: remove this one maybe?

test/starknet/Space.test.ts
```diff
@@ -5,6 +5,12 @@ import { utils } from '@snapshot-labs/sx';
 import { vanillaSetup } from '../shared/setup';
 import { PROPOSE_SELECTOR, VOTE_SELECTOR } from '../shared/constants';
 
+import { hash } from 'starknet';
+const { getSelectorFromName } = hash;
+
+const P = BigInt(getSelectorFromName('__execute__'));
```
pscott: maybe we can find a better name? :)

test/starknet/SpaceFactory.test.ts
```diff
@@ -69,26 +69,26 @@ describe('Space Deployment Testing', () => {
...
+      quorum: quorum,
+      voting_strategy_params_flat: votingStrategyParamsFlat,
+      voting_strategies: votingStrategies,
+      authenticators: authenticators,
+      executors: executors,
     });
     const receipt = await starknet.getTransactionReceipt(txHash);
-    const decodedEvents = await spaceDeployer.decodeEvents(receipt.events);
-
+    // Removing first event as thats from the account contract deployment
```
pscott: that's *

pscott (changes requested):
Good stuff overall! Have left a couple of comments :)

I'm glad we're moving to a "Space as Account" format, I think it looks nice :D!

contracts/starknet/lib/voting_library.cairo
Orland0x: its from the OZ extensibility [standard](https://github.com/OpenZeppelin/cairo-contracts/blob/main/docs/Extensibility.md): 
"Must prefix storage functions with the name of the namespace to prevent clashing with other libraries (e.g. ERC20_balances)"

contracts/starknet/lib/voting_library.cairo
Orland0x: Isnt https://github.com/snapshot-labs/sx-core/blob/b8d4e6adb29c4b8e902e9411f8a7654ccb9dd4fc/contracts/starknet/lib/voting_library.cairo#L684 handling that?

contracts/starknet/lib/voting_library.cairo
Orland0x: did you mean like eg using a solana and an ethereum address in the same proposal?

contracts/starknet/SpaceAccount.cairo
Orland0x: Yeah me and @fabien discussed this. To remove it we would need to modify the account contract a fair bit (which we should probably do). Keeping it for now provides an easy safety switch for the controller as you can just drain the account if something goes wrong with a proposal.

We can talk more about how best to handle it

test/starknet/Space.test.ts
Orland0x: yeah this is just some random code i forgot to delete lol

contracts/starknet/lib/voting_library.cairo
pscott: Yeah ok should've clicked on the link in the PR description ;)
It does give some weird variable names: `Voting_voting_delay_store`, `Voting_min_voting_duration_store`... it does look verbose. But if that's the standard, then we should definitely follow it :)

contracts/starknet/lib/voting_library.cairo
Orland0x: isnt it a uint256 so cant be negative?

pscott: so for `uints` that is correct: but for `felts`, I believe they are signed by default? Which means we could maybe have a negative `min_voting_duration`, or a negative `voting_strategies_len` ?

contracts/starknet/lib/voting_library.cairo
pscott: `cancel_proposal` and `finalize_proposal` are similar except it doesn't perform as much checks and is only callable by `controller`. So as a rule of thumb, any changes made to `finalize_proposal` should probably get reflected into `cancel_proposal` too (rule of thumb, not exactly true).

I don't know if it's clearer now why we'd need to handle `proposal.executor == 1` in `cancel_proposal` too? :)

contracts/starknet/lib/voting_library.cairo
Orland0x: yes I agree we need to think more about invalid cases and failure messages. I created an issue https://github.com/snapshot-labs/sx-core/issues/255

contracts/starknet/lib/voting_library.cairo
Orland0x: Ah i see yes, was looking at the wrong code. will update now

pscott (approved):
LGTM!

PR #258, opened 2022-08-04
Add metadata uri to eth sig

### LINKED ISSUE #202: Add `metadataURI` to the `Propose` type
Currently, `EthSig` does not support `string` types. We need to support those in order to add `metadataURI` to the EIP-712 message data :)

pscott:
Adds metadata URI (and support for any string) to the EIP-712 object used to authenticate users.

Basically, the idea is to use `IntsSequence` (represented in LE, not in BE).
On the Cairo side of things:
- Use `keccak_bigend` to hash the input
- Add `metadata_uri_string_len` as an argument to the `propose` function.

On the TS side of things (offchain):
- Add `LEFromString` method to `IntsSequence` (https://github.com/snapshot-labs/sx.js/pull/48)
- Change `metadata_uri` to be an `IntsSequence`

Closes #202

### DIFF
--- a/contracts/starknet/Authenticators/EthSig.cairo
+++ b/contracts/starknet/Authenticators/EthSig.cairo
@@ -188,6 +188,7 @@ func keccak_ints_sequence{range_check_ptr, bitwise_ptr : BitwiseBuiltin*, keccak
     return keccak_bigend(inputs=sequence, n_bytes=nbytes)
 end
 
+@external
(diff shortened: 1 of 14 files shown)


pscott:
Need to merge https://github.com/snapshot-labs/sx.js/pull/48 and update the version on this PR to have it work I think :)

Orland0x (approved):
LGTM!

PR #90, opened 2022-04-27
Clean some code

pscott:
- Use asserts instead of ifs
- Optimize the step counts in the registration functions (I ran the numbers: https://docs.google.com/spreadsheets/d/1ofjHWUX5VTR7uq-roB-pXv7dnevBN5Nh13gNt8vJu4Q/edit#gid=0)

Closes #85 
Closes #98

PR #264, opened 2022-08-05
Use better asserts in voting lib

### LINKED ISSUE #85: Clean some code
Here's a list of things that can be cleaned:

https://github.com/snapshot-labs/sx-core/blob/0f81eddad82008762e4e6faf0205f337fdb0481f/contracts/starknet/space/space.cairo#L331-L336

https://github.com/snapshot-labs/sx-core/blob/0f81eddad82008762e4e6faf0205f337fdb0481f/contracts/starknet/space/space.cairo#L283-L288

pscott:
Supersedes #90 and #206 
Closes #85


Orland0x:
looks good, found one more one: 
https://github.com/snapshot-labs/sx-core/blob/1551fbd2e2ee44220aa61af1a79bfcdd64bedc82/contracts/starknet/lib/voting.cairo#L440

If proposal doesnt exist then the vote will fail with `"Voting has not started yet"` so we should add an assert not zero

Orland0x (approved):

PR #269, opened 2022-08-08
Fix frontrunning signature

### LINKED ISSUE #204: Avoid meta-tx to be front ran with incorrect parameters
Currently, the types signed are:

```
export const proposeTypes = {
  Propose: [
    { name: 'salt', type: 'uint256' },
    { name: 'space', type: 'bytes32' },
    { name: 'executionHash', type: 'bytes32' },
    // { name: 'metadataURI', type: 'string' }, TODO: Support `string` type in the future
  ],
};

export const voteTypes = {
  Vote: [
    { name: 'salt', type: 'uint256' },
    { name: 'space', type: 'bytes32' },
    { name: 'proposal', type: 'uint256' },
    { name: 'choice', type: 'uint256' },
  ],
};
```

but this means that a user could get frontrun because a lot of data is not currently being signed (`used_voting_strategies`, `user_voting_strategies_params` ...)

pscott:
- Modified the `hash_array` function to add the length *inside* the function (with an alloc) instead of appending it on the input (indeed, appending data at the end of an array that you are not the OWNER of is not a good practice, and was failing when using the array from calldata)
- Extend the EIP-712 signature arguments for the EthSig auth: the signature now signs every argument in the calldata. The reasoning behind this decision is that even though some fields would be possible to omit today, signing everything makes if future proof. We don't need to worry about a future PR that could break the security assumptions regarding front running, since we sign all the data that the user is sending to the SC.

Closes #204


contracts/starknet/Authenticators/EthSig.cairo
```diff
@@ -75,7 +75,7 @@ func prepend_prefix_2bytes{range_check_ptr, bitwise_ptr : BitwiseBuiltin*}(
     prefix : felt, output : Uint256*, input_len : felt, input : Uint256*
 ):
     if input_len == 0:
-        # Done, simlpy store the prefix in the `.high` part of the last Uin256, and
+        # Done, simlpy store the prefix in the `.high` part of the last Uint256, and
```
Orland0x, line 78: simply

contracts/starknet/lib/hash_array.cairo
```diff
@@ -1,13 +1,20 @@
...
-    # Appending the length of the array to itself as the offchain version of the hash works this way
-    assert array[array_len] = array_len
+    # Hash the array
     let (hash_state_ptr) = hash_init()
-    let (hash_state_ptr) = hash_update{hash_ptr=pedersen_ptr}(hash_state_ptr, array, array_len + 1)
+    let (hash_state_ptr) = hash_update{hash_ptr=pedersen_ptr}(hash_state_ptr, array, array_len)
+
+    # Append the length of the array to itself as the offchain version of the hash works this way
+    let (suffix : felt*) = alloc()
+    assert suffix[0] = array_len
```
Orland0x: Rather than allocating a new array, might be cheaper to access the pointer directly. Like:

```
    let (__fp__, _) = get_fp_and_pc()

    let (hash_state_ptr) = hash_update{hash_ptr=pedersen_ptr}(hash_state_ptr, &array_len, 1)
```

pscott: The problem I had was with the `&array_len`. Does `get_fp_and_pc()` fix that?

Orland0x: yep, it allows you to access the pointer

pscott: I'm surprised this works! By querying for `fp` I suddenly can take the address of `array_len`. Wild!
https://github.com/snapshot-labs/sx-core/pull/269/commits/94d07e0d867327656f6c96ddd6c72edf7c3b937d

Orland0x (approved):

Issue #116, opened 2022-05-09
Store voting strategies at the proposal level. 


Orland0x:

pscott:
The idea is to store voting strategies when a proposal is created so that if voting strategies of the space change while the proposal is still ongoing, the proposal still has access to its original voting strategies.
As it's currently hard to stored mapping / arrays inside of structs currently in Cairo, we decided to leave this as it is and come back maybe later on (hence this issue).

pscott:
temporarily fixed by https://github.com/snapshot-labs/sx-core/pull/279 :) 
Leaving this open for the future

zzuziak:
Hey @pscott , is that something that can be implemented for SOC? Will it be possible with Cairo 1.0?

pscott:
> Hey @pscott , is that something that can be implemented for SOC? Will it be possible with Cairo 1.0?

Waiting for Cairo 1.0 to know more about it. Should be manageable but will make everything so much more expensive until we have a Data Availbility Solution so I'm not sure we'll want to implement it straight away...

As for SOC I believe it's the same argument: technically possible but will make proposals more expensive

PR #279, opened 2022-08-13
Update controller tests; Prevent setting modification when a proposal is still active

### LINKED ISSUE #199: Full coverage on controller actions test

### LINKED ISSUE #275: Add test for quorum
We don't have any test to check for `Quorum has not been reached` message :)

pscott:
This PR does two things:
- Updates the controller action tests, making sure it's working as intended, adding some sweet regression tests
- Prevents settings to be modified when a proposal is still active (temporary fix to #116 ) and adds some new tests to check it :)

Replaces #208 (I used @Orland0x 's branch as a base branch)
Closes #199 
Closes #275


test/starknet/ActiveProposal.test.ts
```diff
@@ -0,0 +1,107 @@
+import { expect } from 'chai';
+import { starknet, ethers } from 'hardhat';
+import { vanillaSetup } from '../shared/setup';
+import { StarknetContract, Account } from 'hardhat/types';
+import { utils } from '@snapshot-labs/sx';
+import { getProposeCalldata } from '@snapshot-labs/sx/dist/utils/encoding';
```
Orland0x: can just access from `utils.encoding.getProposeCalldata`

Orland0x:
Most settings are stored at the proposal level, so they can be changed even if a proposal is active right? whats the motivation to prevent that?

pscott:
> Most settings are stored at the proposal level, so they can be changed even if a proposal is active right? whats the motivation to prevent that?

I'm not sure I understand the question. For proposal level settings, we don't need to prevent the modification, since they're already stored at the proposal level. Modifying those will only impact the following proposals, not those already active :)

Orland0x:
> > Most settings are stored at the proposal level, so they can be changed even if a proposal is active right? whats the motivation to prevent that?
> 
> I'm not sure I understand the question. For proposal level settings, we don't need to prevent the modification, since they're already stored at the proposal level. Modifying those will only impact the following proposals, not those already active :)

eg adding an executor, we prevent that despite executors being stored at proposal level

Orland0x (approved):

PR #287, opened 2022-08-17
Voting strategy indexing

### LINKED ISSUE #163: Handling multiple instances of the same strategy but with different parameters
We index voting strategies by their address which means that a space cannot have multiple instances of the same strategy active at once (eg this would be the case for using the aggregate balance of multiple ERC20 tokens to calculate voting power, where multiple single slot strategies would be needed).

One solution to this is to add an additional index to the voting strategies storage such that one indexes each strategy instance by its address and its index rather than just the address. However this adds some complexity to the space contract and overhead in the common case where the feature is not needed. 

 A simpler solution to this is to deploy multiple instances (proxies) of a particular strategy so that each one can be uniquely indexed in the space by its address. We are using this approach for the time being.

Orland0x:
Instead of indexing by voting strategy address, we now index by a  unique number instead. The benefit of this is that it allows a space to register multiple instances of the same strategy (but with different parameters) 

closes https://github.com/snapshot-labs/sx-core/issues/163


test/starknet/ControllerActions.test.ts
```diff
@@ -149,7 +149,7 @@ describe('Controller Actions', () => {
       user.address,
       utils.intsSequence.IntsSequence.LEFromString(''),
       vanillaExecutionStrategy.address,
-      [vanillaVotingStrategy.address],
+      ['0x2'],
```
pscott, line 152: looking at those `0x0`, `0x1`, `0x2`, it might be a good idea to declare a variable `latestIndex` and increment it and use it when need be? Might make working with this code easier in the future

pscott (approved):
Minor improvement maybe we could add in `ControllerActions.test.ts` but other that that it looks good!

test/starknet/ControllerActions.test.ts
Orland0x: feels like a lot of effort right now lol, will defo do that for the starknet.js scripts

PR #295, opened 2022-08-23
Fix eth sig failure

### LINKED ISSUE #292: fix eth signature issue

pscott:
After investigation, I found that calling `.address` in starknet.js would actually left pad the address with some zeroes.
In the cairo code however, I was right-padding with zeroes (as specified in EIP-712....).

This PR removes the right-padding done on the `space` address in the cairo code.
The `proposer_address` / `voter_address` are not padded because they are 20 bytes long and effectively right padded when constructing the EIP-712 hash in js... :)

Closes #292

### DIFF
--- a/contracts/starknet/Authenticators/EthSig.cairo
+++ b/contracts/starknet/Authenticators/EthSig.cairo
@@ -227,7 +227,6 @@ func authenticate_proposal{
 
     # Space address
     let (space) = FeltUtils.felt_to_uint256(target)
-    let (padded_space) = pad_right(space)
 
     # Proposer address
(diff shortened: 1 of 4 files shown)


pscott:
This PR also brings back the [tweak to `hash_array`](https://github.com/snapshot-labs/sx-core/pull/295/commits/f7c95dc8477a156338ad5d58267e13d6b335627b) that was removed in a [previous PR](https://github.com/snapshot-labs/sx-core/commit/3a707cab3811729a7028f96e5127666354db771f) :p

Orland0x (approved):

Orland0x:
LGTM

PR #296, opened 2022-08-23
Session key auth

### LINKED ISSUE #254: Use session keys to broadcast vote and propose
### Motivation
Reduce gas costs to vote in Ethereum DAOs by performing the expensive Ethereum signature verification only once per session. All other votes made by the user during the session are authenticated by a Stark signature with a key stored in the browser. 

### Implementation 
Will be implemented as an authenticator contract called `EthSigSessionKey.cairo`. Spaces will be able to whitelist this authenticator for members of the space to use. The session duration is variable and can be set at deployment inside the constructor. 

### Interface 
```
@constructor 
func constructor(session_length : felt)
end

# Returns session key if it exists and hasn't expired
func get_session_key(eth_address: felt)  -> (key: felt):
end

# Performs EC recover on the Ethereum signature and stores the session key in a 
# mapping indexed by the recovered Ethereum address
@external 
func generate_session_key_from_sig(
    r : Uint256,
    s : Uint256,
    v : felt,
    salt : Uint256,
    session_public_key: felt
):
end

# Checks signature is valid and if so, removes session key for user
@external 
func revoke_session_key(sig_len: felt, sig: felt*):
end

# Calls get_session_key with the ethereum address (calldata[0]) to check that a session is active.
# If so, perfoms stark signature verification to check the sig is valid. If so calls execute with the payload. 
@external
func authenticate(sig_len: felt, sig: felt*, target: felt, function_selector: felt, calldata_len: felt, calldata: felt*)
end
```

Orland0x:
Adding 2 session key authenticators, one authorized by an ethereum signature, and one authorized by an ethereum transaction. 

closes https://github.com/snapshot-labs/sx-core/issues/254


contracts/starknet/Authenticators/EthSigSessionKey.cairo
```diff
@@ -0,0 +1,104 @@
...
+    SessionKey.register_session_key(eth_address, session_public_key, session_duration)
+    return ()
+end
+
+# Checks signature is valid and if so, removes session key for user
+@external
+func revoke_session_key{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(
+    sig_len : felt, sig : felt*, session_public_key : felt
+):
+    # TO DO: sig verification
```
pscott: any blocker here?

contracts/starknet/Authenticators/EthTx.cairo
```diff
@@ -47,25 +20,27 @@ func authenticate{syscall_ptr : felt*, range_check_ptr, pedersen_ptr : HashBuilt
...
-        assert_not_equal(address.value, 0)
-    end
-    # The sender of the commit on L1 must be the same as the voter/proposer L1 address in the calldata.
-    with_attr error_message("Commit made by invalid L1 address"):
-        assert calldata[0] = address.value
-    end
-    # Clear the hash from the contract by writing the zero address to the mapping.
-    commit_store.write(hash, Address(0))
+
+    # Checks that hash maches a commit and that the commit was created by the correct address
```
pscott: matches*

contracts/starknet/Authenticators/EthTxSessionKey.cairo
```diff
@@ -0,0 +1,108 @@
...
+
+    # Check user's address is equal to the owner of the session key
+    with_attr error_message("Invalid Ethereum address"):
+        assert calldata[0] = eth_address
+    end
+
+    # Check signature with session key
+    if function_selector == PROPOSAL_SELECTOR:
+        StarkEIP191.verify_propose_sig(
+            r, s, salt, target, calldata_len, calldata, session_public_key
```
pscott, line 51: No way we can bypass the signature by having `r = 0`, `s = 0` and `eth_address = 0` right? Maybe I'm paranoid but I'd rather be sure

contracts/starknet/Authenticators/EthTxSessionKey.cairo
```diff
@@ -0,0 +1,108 @@
...
+}(eth_address : felt, session_public_key : felt, session_duration : felt):
+    alloc_locals
+    # Cast arguments to single array and hash them
+    let (input_array : felt*) = alloc()
+    assert input_array[0] = eth_address
+    assert input_array[1] = session_public_key
+    assert input_array[2] = session_duration
+    let (hash) = HashArray.hash_array(3, input_array)
+
+    # Checks that hash maches a commit and that the commit was created by the correct address
```
pscott: matches*

contracts/starknet/lib/eth_tx.cairo
```diff
@@ -0,0 +1,54 @@
...
+        with_attr error_message("Invalid message origin address"):
+            assert from_address = origin
+        end
+        # Note: If the same hash is committed twice by the same sender, then the mapping will be overwritten but with the same value as before.
+        EthTx_commit_store.write(hash, sender)
+        return ()
+    end
+
+    # Checks to see if commit exists, if so clears it from the contract, else throws
+    func check_commit{syscall_ptr : felt*, range_check_ptr, pedersen_ptr : HashBuiltin*}(
```
pscott: I would suggest we use another name than `check` because intuitively, `check` only `checks` something, whereas here we check and then overwrite it.

I was thinking about:
- `consume_commit`
- `pop_commit`
- `use_commit`

I personally like `consume_commit` more, wdyt?

contracts/starknet/Authenticators/EthTxSessionKey.cairo
```diff
@@ -0,0 +1,108 @@
...
+        end
+    end
+
+    # Call the contract
+    execute(target, function_selector, calldata_len, calldata)
+
+    return ()
+end
+
+# Performs EC recover on the Ethereum signature and stores the session key in a
```
pscott: We're not performing any EC recover in the current code but simply matching the `hash`.

If we can be truly certain that `sender` (in `EthTx.commit`) is indeed the `msg.sender` (on l1), then I guess we don't need to do any EC recover, right?

(I think we can because `msg.sender` is what's used in `StarkNetCommit.sol` :p)

contracts/starknet/lib/session_key.cairo
```diff
@@ -0,0 +1,61 @@
...
+from starkware.starknet.common.syscalls import get_block_timestamp
+from starkware.cairo.common.cairo_builtins import HashBuiltin
+from starkware.cairo.common.math import assert_le, assert_not_zero
+
+@storage_var
+func SessionKey_owner(session_public_key : felt) -> (eth_address : felt):
+end
+
+@storage_var
+func SessionKey_end_timestamp(session_public_key : felt) -> (timestamp : felt):
```
pscott: missing a `_store` suffix?

contracts/starknet/lib/session_key.cairo
```diff
@@ -0,0 +1,61 @@
+%lang starknet
+
+from starkware.starknet.common.syscalls import get_block_timestamp
+from starkware.cairo.common.cairo_builtins import HashBuiltin
+from starkware.cairo.common.math import assert_le, assert_not_zero
+
+@storage_var
+func SessionKey_owner(session_public_key : felt) -> (eth_address : felt):
```
pscott: missing a `_store` suffix?

contracts/starknet/lib/session_key.cairo
```diff
@@ -0,0 +1,61 @@
...
+        let (current_timestamp) = get_block_timestamp()
+        SessionKey_end_timestamp.write(session_public_key, current_timestamp + session_duration)
+        session_key_registered.emit(eth_address, session_public_key, session_duration)
+        return ()
+    end
+
+    func revoke_session_key{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(
+        sig_len : felt, sig : felt*, session_public_key : felt
+    ):
+        # TODO: Verify stark signature
```
pscott: O_o

contracts/starknet/lib/session_key.cairo
```diff
@@ -0,0 +1,61 @@
...
+
+        # Set session key owner to zero
+        SessionKey_owner.write(session_public_key, 0)
+        session_key_revoked.emit(session_public_key)
+        return ()
+    end
+
+    # Returns owner of a session key if it exists, otherwise returns 0
```
pscott: It does not return `0` otherwise but currently throws :) might be better to actually make it return `0` though!

contracts/starknet/lib/stark_eip191.cairo
```diff
@@ -94,12 +104,12 @@ namespace StarkSig:
 
         assert message[0] = STARKNET_MESSAGE
         assert message[1] = DOMAIN_HASH
-        assert message[2] = proposer_address
+        assert message[2] = authenticator
```
pscott, line 109: why the switch to authenticator rather than proposer?

scripts/deploySpace1.ts
```diff
@@ -0,0 +1,106 @@
+import fs from 'fs';
```
pscott: Why `deploySpace1.ts`? Could we find a more explicit name maybe?

scripts/test-crosschain.sh
```diff
@@ -4,7 +4,10 @@ yarn wait-on tcp:8000 &&
 wait-on tcp:8545 && 
 yarn hardhat test test/crosschain/ZodiacExecution.test.ts --network 'ethereumLocal' --starknet-network 'starknetLocal' && 
 sleep 10 && 
-yarn hardhat test test/crosschain/EthTxAuth.test.ts --network 'ethereumLocal' --starknet-network 'starknetLocal'
+yarn hardhat test test/crosschain/EthTxAuth.test.ts --network 'ethereumLocal' --starknet-network 'starknetLocal' && 
+sleep 10 &&
```
pscott, line 8: why do we add a sleep here?

test/crosschain/EthTxSessionKeyAuth.test.ts
```diff
@@ -0,0 +1,214 @@
...
+  it('should authorize a session key from an L1 transaction and allow authentication via it', async () => {
+    await starknet.devnet.loadL1MessagingContract(networkUrl, mockStarknetMessaging.address);
+    // Committing the hash of the payload to the StarkNet Commit L1 contract
+    await starknetCommit
+      .connect(account)
+      .commit(
+        ethTxSessionKeyAuth.address,
+        getCommit(ethAddress, sessionPublicKey, sessionDuration)
+      );
+    // Checking that the L1 -> L2 message has been propogated
```
pscott: propagated*

pscott (changes requested):
Good code! A couple of changes requested but should be fine!

contracts/starknet/Authenticators/EthSigSessionKey.cairo
Orland0x: added

contracts/starknet/Authenticators/EthTxSessionKey.cairo
Orland0x: this line would fail then: `verify_ecdsa_signature(message_hash, public_key, r, s)` right?

Orland0x: we can add more extensive edge case testing to the sig verifications libs if you think its necessary

contracts/starknet/lib/eth_tx.cairo
Orland0x: yeah agree. ill like consume too. updated

contracts/starknet/Authenticators/EthTxSessionKey.cairo
Orland0x: yeah just copy and pasted some code and forgot to delete comments

contracts/starknet/lib/session_key.cairo
Orland0x: na i think throwing is better actually, so you get the right error message. Otherwise there is no way to know why it failed (either expired or never initialized). I updated the comment

contracts/starknet/lib/stark_eip191.cairo
Orland0x: I think the eip191 standard sets the data out this way: https://eips.ethereum.org/EIPS/eip-191. 

ie you add the verifying address in:

```
    // Arguments when calculating hash to validate
    // 1: byte(0x19) - the initial 0x19 byte
    // 2: byte(0) - the version byte
    // 3: this - the validator address
    // 4-7 : Application specific data
```

scripts/test-crosschain.sh
Orland0x: some issues with the devnet, might be working now with the updates

scripts/deploySpace1.ts
Orland0x: thats an old file, deleted it

contracts/starknet/Authenticators/EthTxSessionKey.cairo
pscott: > this line would fail then: `verify_ecdsa_signature(message_hash, public_key, r, s)` right?

that is indeed what I was thinking but I just want to make sure. It would be a shame if we have an exploit just because we `assumed` that this function behaves like that (maybe it has a special case for 0? dunno)

contracts/starknet/lib/stark_eip191.cairo
pscott: Ok yup sounds good :)

contracts/starknet/lib/session_key.cairo
```diff
@@ -0,0 +1,159 @@
...
+        return (eth_address)
+    end
+end
+
+func register{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*, range_check_ptr}(
+    eth_address : felt, session_public_key : felt, session_duration : felt
+):
+    SessionKey_owner_store.write(session_public_key, eth_address)
+    let (current_timestamp) = get_block_timestamp()
+    SessionKey_end_timestamp_store.write(session_public_key, current_timestamp + session_duration)
```
pscott, line 152: Should we check for overflow here? What happens if `current_timestamp + session_duration` overflows? the number will be either negative, or `< current_timestamp`? This means it simply won't work, so might as well throw an error here right?

pscott (approved):
Just one question about overflow but otherwise looks good! Very nice work, I want to try it with a UI! :)

contracts/starknet/lib/session_key.cairo
Orland0x: yeah nice spot. could also enforce a max session duration?

pscott: Yeah I think a max duration of a year should be fine to start with, wdyt?

Orland0x: I just put an overflow check for now, some people might just want an indefinite session key to be active so seems unnecessary to limit it.

```   
with_attr error_message("Overflow in Session duration, use smaller value"):
        assert_le(current_timestamp, end_timestamp)
    end
```

pscott (approved):

PR #299, opened 2022-08-24
Fix infinite vp

### LINKED ISSUE #298: Add check to ensure each voting can only be used once.
Currently you can pass the same voting strategy multiple times and increase your voting power.

pscott:
Adds a `assert_no_duplicates` function that will throw if duplicates are found in an array. This function is doing this the "dumb way" (i.e. for each element, loop through the array and make sure there a no duplicates, resulting in O(n^2) complexity). The good thing about this is that no builtins are used whatsoever: so it's step intensive but builtin lightweight.
Another way we could do this is by using the `bitwise builtin`. Since most spaces will probably not support a gazilion voting strategies, I think we're fine with that, but we could indeed other ways of doing this in the future (sorting by increasing order for example to make lookup faster...)

Call this function in `get_cumulative_voting_power` to make sure voting strategies are unique.

Closes #298

### DIFF
--- a/contracts/starknet/lib/voting.cairo
+++ b/contracts/starknet/lib/voting.cairo
@@ -1026,9 +1026,70 @@ func assert_no_active_proposal{syscall_ptr : felt*, pedersen_ptr : HashBuiltin*,
     return ()
 end
 
+# Asserts that the array does not contain any duplicates.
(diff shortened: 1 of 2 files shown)


Orland0x (approved):
LGTM!

PR #305, opened 2022-08-28
Add Starknet signature

### LINKED ISSUE #282: Add StarknetSig auth
Similarly to EthereumSigAuth, it would be nice to have the Starknet Sig auth :)

pscott:
- Adds Starknet Signature Authenticator
- Adds corresponding tests 

Implem details:
- `metadataURI` had to be split into the `IntsSequence` type
- This authenticator does not support Argent / OZ (by that I mean, regular account contracts) signatures (indeed it does not call `Account.is_valid_signature`). We need to decide on whether we want a different authenticator for those accounts, or if we wish to simply pass an argument like `is_wallet` to the auth

Closes #282


Orland0x (approved):

Orland0x:
LGTM

PR #306, opened 2022-08-29
Add auth to ethsig

pscott:
- Adds the `authenticator` parameter to the EthSig.cairo authenticator
- Also adds `chainId` to the domain


Orland0x (approved):
LGTM

PR #310, opened 2022-08-30
increment counter BEFORE recursive call

Orland0x:


contracts/starknet/lib/voting.cairo
```diff
@@ -880,9 +880,10 @@ func unchecked_add_voting_strategies{
         # The following elements are the actual params
         unchecked_add_voting_strategy_params(next_index, 1, params_len, params)
 
+        Voting_num_voting_strategies_store.write(next_index + 1)
+
```
pscott: So now that I think about it, even though starknet should, in theory, only count a single `write` operation even if we do multiple ones (because they are modifying the same address), currently that's not the case.
I think we should pass `next_index` as a parameter to this function, and wrap this function inside another one that does something like:
```
read the index store
call wrapped function to add strategies
write the new index (should be `prev_index + addresses_len`)
```

pscott (changes requested):
If we're at it, I think we might as well optimise the function, wdyt?

contracts/starknet/lib/voting.cairo
Orland0x: The same thought crossed my mind but I cba at the time lol. will do though

contracts/starknet/lib/voting.cairo
```diff
@@ -221,9 +221,13 @@ namespace Voting:
             voting_strategy_params_flat_len, voting_strategy_params_flat
         )
 
+        let (prev_index) = Voting_num_voting_strategies_store.read()
         unchecked_add_voting_strategies(
-            voting_strategies_len, voting_strategies, voting_strategy_params_all, 0
+            voting_strategies_len, voting_strategies, voting_strategy_params_all, prev_index, 0
         )
+        # Incrementing the voting strategies counter by the number of strategies added
+        Voting_num_voting_strategies_store.write(prev_index + voting_strategies_len)
```
pscott: eeeh maybe I'm a bit picky but I'd rather have a wrapper function that does all this logic rather than having to manage the `Voting_num_voting_strategies_store` outside of `unchecked_add_voting_strategies`.

I think `unchecked_add_voting_strategies` should do this logic, and maybe have a wrapper inside if (something like `unchecked_add_voting_strategies_recurse`) to handle the recursion. Wdyt?

pscott (changes requested):

contracts/starknet/lib/voting.cairo
Orland0x: yep agree

pscott:
Eh I think you added some `abi` file to the PR ^^'

Orland0x:
> Eh I think you added some `abi` file to the PR ^^'

oops, deleted

pscott (approved):
LGTM!

PR #311, opened 2022-08-30
Hash array fix

Orland0x:
Missed call to finalize keccak in the hash array lib function. Added this and a test.


contracts/starknet/lib/hash_array.cairo
```diff
@@ -1,5 +1,5 @@
 from starkware.cairo.common.cairo_builtins import HashBuiltin
-from starkware.cairo.common.hash_state import hash_init, hash_update
+from starkware.cairo.common.hash_state import hash_init, hash_update, hash_finalize
 from starkware.cairo.lang.compiler.lib.registers import get_fp_and_pc
```
pscott: i believe this import can be removed :D

pscott (approved):

PR #327, opened 2022-09-09
more test coverage on zodiac module

Orland0x:
Added some more test coverage and some better failure cases.


contracts/ethereum/ZodiacModule/SnapshotXL1Executor.sol
```diff
@@ -211,6 +211,7 @@ contract SnapshotXL1Executor is Module, SnapshotXProposalRelayer {
    * @param executionHash Hash of all the transactions in the proposal
    * @param proposalOutcome Whether proposal was accepted / rejected / cancelled
    * @param _txHashes Array of transaction hashes in proposal
+   * @notice REMEMBER TO REMOVE BEFORE PROD
```
pscott: maybe we should add the `TODO` keyword in case we search for all `TODO` before going to prod :)

pscott (approved):
lgtm :) will edit and merge

PR #328, opened 2022-09-12
multi voting strategy testing

### LINKED ISSUE #302: Add extensive testing of multiple voting strategies on a single proposal
So far all of our tests use just a single strategy, we should add tests to ensure the cumulation is done correctly including tests on overflow, duplicate strategies etc

Orland0x:
Added a test showing correct agggregation of voting power when multiple strategies are used. Also a test to show that the vote tx fails if an invalid voting strategy is submitted. 

closes https://github.com/snapshot-labs/sx-core/issues/302


pscott (approved):
yup, needed! :)

PR #338, opened 2022-09-22
Change chainId to goerli instead of SN_GOERLI

pscott:
MM doesn't allow users to sign with a chainId different than `5` (goerli's chainId) so we need to change the `chainId` parameter we're using in our contracts


Orland0x (approved):

Orland0x:
LGTM

PR #342, opened 2022-09-23
Finalize even if quorum has not been reached

### LINKED ISSUE #339: Proposal can't be finalized if quorum hasn't been reached
If a proposal never reaches quorum, then it can never get finalized. Indeed the function `finalize_proposal` will throw an error ("Quorum has not been reached").

To fix this, I think the logic should be like this:
1. Ensure `min voting period` has elapsed (throw if that's not the case)
2. Check if quorum has been reached
  - 2.1 If quorum has not been reached, check if `max voting period` has elapsed
    - 2.1.1 If `max voting period` has elapsed, continue with 3
       2.1.2 If `max voting period` has not elapsed, throw by saying quorum has not been reached
  - 2.2 If quorum has been reached, continue to 3
3. Carry on with the finalize logic

Also, would be nice to add a regression test for this case :)

pscott:
Fixes #339 !

Also renames `add_executors` / `remove_executors` to `add_execution_strategies` / `remove_execution_strategies`


contracts/starknet/TestContracts/FailsIfRejected.cairo
```diff
@@ -0,0 +1,17 @@
+%lang starknet
```
Orland0x: Maybe add Execution Strategy to the name so its clear what this contract is for

pscott: [a06f653](https://github.com/snapshot-labs/sx-core/pull/342/commits/a06f65378caf367f9ca12567edaa8082f09f8aca)

Orland0x (commented):
LGTM

Orland0x (approved):

PR #355, opened 2022-09-28
removed storage of vote data

### LINKED ISSUE #353: Remove VP and choice storage for each voter
Unecessary data that can be accessed from the indexer.

Orland0x:
closes https://github.com/snapshot-labs/sx-core/issues/353

### DIFF
--- a/contracts/starknet/SpaceAccount.cairo
+++ b/contracts/starknet/SpaceAccount.cairo
@@ -212,10 +212,10 @@ func cancel_proposal{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check
 //
 
 @view
-func get_vote_info{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr: felt}(
-    voter_address: Address, proposal_id: felt
-) -> (vote: Vote) {
-    return Voting.get_vote_info(voter_address, proposal_id);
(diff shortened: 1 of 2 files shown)


pscott (approved):
nice & simple

contracts/starknet/SpaceAccount.cairo
```diff
@@ -212,10 +212,10 @@ func cancel_proposal{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check
 //
 
 @view
-func get_vote_info{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr: felt}(
-    voter_address: Address, proposal_id: felt
-) -> (vote: Vote) {
-    return Voting.get_vote_info(voter_address, proposal_id);
+func has_voted{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr: felt}(
```
bonustrack, line 215: Why we have has_voted function twice? Couldnt we reuse the same function?

pscott: One is external, the other is not. This is to follow OZ's pattern :)

Basically the logic is in a `lib` file with its namespace: and when you want to write `external` functions it's mostly wrappers around this lib. It's done to avoid namespace / storage clashing ;)

PR #356, opened 2022-09-28
add metadata uri to space

### LINKED ISSUE #349: Add "metadata_uri" var at the space level with "set_metadata_uri" to change it

Orland0x:
closes https://github.com/snapshot-labs/sx-core/issues/349

### DIFF
--- a/contracts/starknet/SpaceAccount.cairo
+++ b/contracts/starknet/SpaceAccount.cairo
@@ -34,6 +34,8 @@ func constructor{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr
     authenticators: felt*,
     executors_len: felt,
     executors: felt*,
+    metadata_uri_len: felt,
+    metadata_uri: felt*,
 ) {
     Account.initializer(public_key);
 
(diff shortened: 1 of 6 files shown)


pscott (approved):
All good! :)

contracts/starknet/SpaceAccount.cairo
```diff
@@ -277,6 +281,14 @@ func update_proposal_threshold{
     return ();
 }
 
+@external
+func update_metadata_uri{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr: felt}(
```
bonustrack, line 285: I think it should be called `set_metadata_uri`, following OZ standard for NFT with tokenURI and setTokenURI

pscott: I think Orlando used `update` to go along with with other `update_*` functions we have. Do you think we should rename all `update_*` to `set_*` ?

Orland0x: the rest of the space parameters have `update` rather than `set` functions. why do we need to keep to that standard? doesnt seem relevant here

bonustrack: If we use "update" everywhere we can use it here too, but I recommend move to set everywhere at some point. It's just much shorter, more common.

PR #359, opened 2022-09-30
added prefixes

### LINKED ISSUE #274: Add library prefix in error message
Would be nice to have the library name prefix in the error message (similar to how OZ has `Ownable: ...` error message, maybe Voting lib should have `Voting: ...` ?

wdyt @Orland0x ?

Orland0x:
closes #274


contracts/starknet/VotingStrategies/Whitelist.cairo
```diff
@@ -45,7 +45,7 @@ func register_whitelist{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_ch
         // Add it to the whitelist
         let voting_power = Uint256(_whitelist[1], _whitelist[2]);
 
-        with_attr error_message("Invalid uin256") {
+        with_attr error_message("WhitelistInvalid uint256") {
```
pscott: `Whitelist: Invalid uint256` *

pscott (approved):
small typo, otherwise looks good

PR #361, opened 2022-10-03
Updated interfaces

### LINKED ISSUE #360: Add interface for space contract

Orland0x:
Added a interface for the space and space factory contracts. Removed the `execute` method from the Authenticator interface as its not an external function.

closes #360

### DIFF
--- a/contracts/starknet/Interfaces/IAuthenticator.cairo
+++ b/contracts/starknet/Interfaces/IAuthenticator.cairo
@@ -1,6 +1,0 @@
-%lang starknet
-
-// Authenticator contracts have no standard interface, we leave this here for completeness
-@contract_interface
-namespace IAuthenticator {
-}
--- a/contracts/starknet/Interfaces/IAuthenticator.cairo
+++ b/contracts/starknet/Interfaces/IAuthenticator.cairo
@@ -1,7 +1,6 @@
 %lang starknet
 
+// Authenticator contracts have no standard interface, we leave this here for completeness
 @contract_interface
 namespace IAuthenticator {
-    func execute(target: felt, function_selector: felt, calldata_len: felt, calldata: felt*) {
-    }
 }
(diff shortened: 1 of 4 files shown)


contracts/starknet/Interfaces/IAuthenticator.cairo
```diff
@@ -1,7 +1,6 @@
 %lang starknet
 
+// Authenticator contracts have no standard interface, we leave this here for completeness
 @contract_interface
 namespace IAuthenticator {
-    func execute(target: felt, function_selector: felt, calldata_len: felt, calldata: felt*) {
-    }
```
pscott: I'm not sure why we have this? What's the point of an Interface if there's no function in it?

contracts/starknet/Interfaces/ISpaceAccount.cairo
```diff
@@ -0,0 +1,113 @@
...
+    func getPublicKey() -> (publicKey: felt) {
+    }
+
+    func supportsInterface(interfaceId: felt) -> (success: felt) {
+    }
+
+    func setPublicKey(newPublicKey: felt) {
+    }
+
+    func isValidSignature(hash: felt, signature_len: felt, signature: felt*) -> (isValid: felt) {
```
pscott, line 19: Really sucks to be using `camelCase` and then `snake_case` for `signature_len` and `signature` but I guess there's no way around it. Feels terribly wrong tho hahaha

pscott (approved):

contracts/starknet/Interfaces/IAuthenticator.cairo
Orland0x: Okay ill remove

contracts/starknet/Interfaces/ISpaceAccount.cairo
Orland0x: agreed

PR #362, opened 2022-10-04
Felt math

Orland0x:
Went through mathematical asserts / comparisons  and made a few updates. Also used OZ Uint2656 safemath for voting power aggregation


contracts/starknet/lib/voting.cairo
```diff
@@ -486,8 +479,7 @@ namespace Voting {
 
         // Make sure `choice` is a valid choice
         with_attr error_message("Voting: Invalid choice") {
-            assert_le(Choice.FOR, choice);
-            assert_le(choice, Choice.ABSTAIN);
+            assert (choice - Choice.ABSTAIN) * (choice - Choice.FOR) * (choice - Choice.AGAINST) = 0;
```
Orland0x, line 482: this is 2.5x more gas efficient than the 2 `assert_le` comparisons.

pscott: nice :p
I really need to get used to dealing with prime fields :D

contracts/starknet/lib/voting.cairo
```diff
@@ -505,20 +497,16 @@ namespace Voting {
...
         let (no_voting_power) = uint256_eq(Uint256(0, 0), user_voting_power);
-
         with_attr error_message("Voting: No voting power for user") {
             assert no_voting_power = 0;
         }
 
         let (previous_voting_power) = Voting_vote_power_store.read(proposal_id, choice);
-        let (new_voting_power, overflow) = uint256_add(user_voting_power, previous_voting_power);
-
         with_attr error_message("Voting: Overflow in voting power") {
```
pscott, line 505: Can we nest `error_message`s ?
Because the library already has an error message: https://github.com/OpenZeppelin/cairo-contracts/blob/893f4b98739bdc1b7d64d0511d96d19f290eef88/src/openzeppelin/security/safemath/library.cairo#L29-L31

pscott (approved):

pscott (approved):
Should we be using `assert_le_felt` (i.e using treating the input as `unsigned` instead of `signed` numbers)?

Orland0x:
> Should we be using `assert_le_felt` (i.e using treating the input as `unsigned` instead of `signed` numbers)?

resolved on discord

contracts/starknet/lib/voting.cairo
Orland0x: yeah you can nest them. the outer one isnt needed It just makes it easier to see where the error occurs i thought.

pscott:
I think we can squash and merge then :)

PR #363, opened 2022-10-05
Pack proposal struct

### LINKED ISSUE #184: Felt packing inside Proposal Struct
Currently there are 4 timestamps stored inside the proposal struct. 32 bits is sufficient for each timestamp and a felt is 251. Therefore we can pack them all inside a single felt to reduce the storage requirements and therefore gas costs. 

Low Priority as this adds complexity and will not reduce cost per vote.

Orland0x:
closes #184

Reduces cost to create a proposal by nearly 30% for the vanilla setup


contracts/starknet/lib/felt_utils.cairo
```diff
@@ -57,4 +28,32 @@ namespace FeltUtils {
...
+
+    // Packs 4 32 bit numbers into a single felt
+    func pack_4_32_bit{range_check_ptr}(num1: felt, num2: felt, num3: felt, num4: felt) -> (
+        packed_felt: felt
+    ) {
+        with_attr error_message("FeltUtils: number too big to be packed") {
+            assert_lt(num1, SHIFT_32);
+            assert_lt(num2, SHIFT_32);
+            assert_lt(num3, SHIFT_32);
+            assert_lt(num4, SHIFT_32);
```
pscott: Should it be `assert_nn_lt`? If the number is `-1` we want to error, right?

contracts/starknet/lib/voting.cairo
```diff
@@ -458,24 +464,27 @@ namespace Voting {
         }
 
         let (proposal) = Voting_proposal_registry_store.read(proposal_id);
+
+        // Unpacking the the timestamps from the packed value
```
pscott: `the the`

contracts/starknet/lib/proposal.cairo
```diff
@@ -2,10 +2,9 @@ from starkware.cairo.common.uint256 import Uint256
 
 struct Proposal {
     quorum: Uint256,
-    snapshot_timestamp: felt,
-    start_timestamp: felt,
-    min_end_timestamp: felt,
-    max_end_timestamp: felt,
+    // timestamps contains the following packed into a single felt (each one is 32 bit):
+    // snapshot_timestamp, start_timestamp, min_end_timestamp, max_end_timestamp
+    timestamps: felt,
```
pscott, line 7: can we use a `TyepAlias` here maybe?

Orland0x: how do you mean?

pscott: the type here is `felt` but we can use `TypeAlias` ([see here](https://www.cairo-lang.org/docs/how_cairo_works/types.html#user-defined-type-aliases)) to have something like:
```
using PackedTimestamps = (inner: felt)
```
and then we could use `timestamps: PackedTimestamps`

this way the reader understands that it's a "special" felt :)

Orland0x: something like `PackedTimestamps = (packed_timestamps : felt);` seems a little unnecessary. Its an internal value that users wont ever need to interact with directly anyway so i think its okay to keep it a felt.

pscott: Your choice :) I guess it's just the Rust way of thinking that got me wanting this. I find it much cleaner to have a type wrapper for this `packedTimestamp`. If I'm reading the code and I see this `felt` I will be thinking "what the fuck?". If I read the code and I see it's a special type (that just wraps a felt, but it doesn't matter), I will understand more what it is. But honestly it's a detail :)

Orland0x: I just tried and we would need to do all sorts of casting to make it work as the lib uses felts. It would make more sense to have a generic type in the FeltUtils lib, a type like `4x32BitPacked` . then use `timestamps: 4x32BitPacked` in the proposal struct.

Orland0x: Okay tried doing this and I cant get it to compile. I dont think type aliases can be used for return values

Orland0x: lets just stick to felts for the time being I think

pscott: agreed

pscott (approved):

PR #384, opened 2022-11-14
Add merkle whitelist voting strategy 

### LINKED ISSUE #382: Create Whitelist Voting Strategy using a merkle tree
Will enable large whitelists to be used efficiently within SX

Orland0x:
closes https://github.com/snapshot-labs/sx-core/issues/382

### DIFF
--- a/test/starknet/MerkleProof.test.ts
+++ b/test/starknet/MerkleProof.test.ts
@@ -1,66 +1,8 @@
 import { StarknetContract } from 'hardhat/types/runtime';
 import { expect } from 'chai';
 import { ethers, starknet } from 'hardhat';
-import { computeHashOnElements, pedersen } from 'starknet/dist/utils/hash';
-
-class Merkle {
-  root: string;
-  constructor(values: string[]) {
-    this.root = generateMerkleRoot(values);
-  }
-
-  getProof(values: string[], index: number): string[] {
-    return getProofHelper(values, index, []);
-  }
-}
-
-function generateMerkleRoot(values: string[]): string {
-  if (values.length == 1) {
-    return values[0];
-  }
-  if (values.length % 2 != 0) {
-    values.push('0x0');
-  }
-  const nextLevel = getNextLevel(values);
-  return generateMerkleRoot(nextLevel);
-}
-
-function getNextLevel(level: string[]): string[] {
-  const nextLevel = [];
-  for (let i = 0; i < level.length; i += 2) {
-    let node = '0x0';
-    if (BigInt(level[i]) < BigInt(level[i + 1])) {
-      node = pedersen([level[i], level[i + 1]]);
-    } else {
-      node = pedersen([level[i + 1], level[i]]);
-    }
-    nextLevel.push(node);
-  }
-  return nextLevel;
-}
-
-function getProofHelper(level: string[], index: number, proof: string[]): string[] {
-  if (level.length == 1) {
-    return proof;
-  }
-  if (level.length % 2 != 0) {
-    level.push('0x0');
-  }
-  const nextLevel = getNextLevel(level);
-  let indexParent = 0;
-
-  for (let i = 0; i < level.length; i++) {
-    if (i == index) {
-      indexParent = Math.floor(i / 2);
-      if (i % 2 == 0) {
-        proof.push(level[index + 1]);
-      } else {
-        proof.push(level[index - 1]);
-      }
-    }
-  }
-  return getProofHelper(nextLevel, indexParent, proof);
-}
+import { computeHashOnElements } from 'starknet/dist/utils/hash';
+import { MerkleTree } from '../shared/merkle';
 
 describe('Merkle:', () => {
   let testMerkle: StarknetContract;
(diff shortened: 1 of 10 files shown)


test/starknet/MerkleProof.test.ts
```diff
@@ -0,0 +1,89 @@
...
+        ethers.utils.hexlify(ethers.utils.randomBytes(1)),
+      ]);
+    }
+
+    // computing the hash of each address value pair, and sorting
+    const merkleData = values
+      .map((v, i) => [computeHashOnElements(v), v[0], v[1]])
+      .sort(function (a, b) {
+        if (a > b) return 1;
+        if (a < b) return -1;
```
pscott, line 33: why are we sorting the values here? this will make the hashes sorted, but will destroy the ordering required to recompute the hashes, won't it?

test/starknet/MerkleProof.test.ts
```diff
@@ -0,0 +1,89 @@
...
+        return 0;
+      })
+      .map((x, i) => [x[0], x[1], x[2], i]);
```
pscott: should there be a function that does this in the library maybe?

test/starknet/MerkleProof.test.ts
```diff
@@ -0,0 +1,89 @@
...
+      await testMerkle.call('testAssertValidLeaf', {
+        root: tree.root,
+        leaf: [leafData[1], leafData[2]],
+        proof: corruptedProof,
+      });
+    } catch (error: any) {
+      expect(error.message).to.contain('Merkle: Invalid proof');
+    }
+  }).timeout(600000);
+});
```
pscott, line 165: I'd add a couple of different tests:
- Empty merkle tree
- Merkle tree with 1 element
- With 2 elements
- With 3 elements
- With 10000 elements (a big one)
- Asking to prove a leaf that is NOT in the tree

test/starknet/MerkleWhitelist.test.ts
```diff
@@ -0,0 +1,125 @@
...
+          address1,
+          power1.low,
+          power1.high,
+          ...tree.getProof(leaves, Number(leaf1[3])),
+        ],
+      });
+    } catch (err: any) {
+      expect(err.message).to.contain('MerkleWhitelist: Invalid proof supplied');
+    }
+  }).timeout(1000000);
```
pscott, line 124: would add test to prove a leaf that is not in the tree

pscott (changes requested):
I think we can add a bit more tests, and I think I'm missing why we need to sort the hashes :)

Also, is there an npm package that can take a `hash_function` as a parameter to build a merkletree?:)

Orland0x:
> I think we can add a bit more tests, and I think I'm missing why we need to sort the hashes :)
> 
> Also, is there an npm package that can take a `hash_function` as a parameter to build a merkletree?:)

If you dont sort the hashes, then you need an additional positional argument for each hash in order to reconstruct the root. So it seems like standard practice to do so. 

Yeah there is a lib I tried to used called `merkletreejs`. I couldn't get it to work with pedersen though. Pedersen is unusual in that it takes 2 numbers and produces the hash whereas most hash functions concatenate the numbers (eg with `abi.encode`) then hash that single number. 

Agreed about the tests - will add more. I basically just copied this implementation, so wasn't too worried about correctness. https://github.com/ncitron/cairo-merkle-distributor

test/starknet/MerkleProof.test.ts
Orland0x: the hash of each leaf is just `computeHashOnElements([address, vp.low, vp.high])`, so sorting doesnt affect them.

test/starknet/MerkleWhitelist.test.ts
Orland0x: proof of non-membership isnt required here though

pscott: Sure! But making sure it fails would be nice :)

test/starknet/MerkleProof.test.ts
pscott: I guess I'm missing something.
The way I view it, we have a tree, and build a list of hashes, from the bottom to the top, of the required hashes in order to recompute the root, given a leaf.
```
        [8]
       /     \
   [6]       [7]
  /     \     /    \
[2] [3] [4] [5]
```

E.g: if I want to prove [3], then the proofs will be [2, 7]. The verifier then simply needs to hash [2, 3] to get [6], and then has [6,7] to get [8].

If we sort the array `[2, 7]`, this won't work anymore. So I guess I'm missing something: or maybe I'm completely misunderstanding what we're even doing here! ^^

Orland0x: sorry missed this. 

Whats being sorted ^ there is the leaf nodes, not the proof. The list of hashes does not include the ones at higher levels (6,7,8 in that example)

test/starknet/MerkleProof.test.ts
Orland0x: I added a test for empty, 1,2,3,1000 

tests for Non-membership proofs are not required (and are more sophisticated) so unnecessary to add here I think

test/starknet/MerkleProof.test.ts
pscott: Ah indeed my bad, I thought I was looking at the `getProof` implementation haha. I understand now the quiproquo.
All fine ser

pscott (approved):
All good ser!

utACK ;) (but CI will test for me :D)

PR #388, opened 2022-12-02
Throw in all try/catch assertions to catch false-positive tests

solimander:
This PR adds a `throw` statement at the end of all `try` blocks found in the tests to ensure that tests cannot pass because the expected error was not thrown.

This caught three false-positives:

* `Packing should fail if a number greater than 32 bits is used` - The test number used did not actually have 36 bits.
* `Reverts when querying an invalid proposal id` - This function is not expected to revert when a proposal ID does not exist.
* `Authentication should fail if the correct hash of the payload is not committed on l1 before execution is called` - The test was using the correct selector.

### DIFF
--- a/contracts/starknet/lib/voting.cairo
+++ b/contracts/starknet/lib/voting.cairo
@@ -799,6 +799,9 @@ namespace Voting {
         proposal_id: felt
     ) -> (proposal_info: ProposalInfo) {
         let (proposal) = Voting_proposal_registry_store.read(proposal_id);
+        with_attr error_message("Voting: Proposal does not exist") {
+            assert_not_zero(proposal.timestamps);
+        }
 
         let (power_against) = Voting_vote_power_store.read(proposal_id, Choice.AGAINST);
         let (power_for) = Voting_vote_power_store.read(proposal_id, Choice.FOR);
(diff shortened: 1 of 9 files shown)


solimander:
Oh, didn't run crosschain tests locally. Looks like one more false-positive... pushing fix now.

pscott (approved):
Thanks for your PR! ❤️ 

I'm glad you caught these. I'm also glad the code itself is behaving as expected, and it's the "only" the tests that invalid.

I'll let @Orland0x have a look before we merge.

One thing that caught my attention is the `reverts when querying an invalid proposal id`. I'm wondering if it would make more sense, for the end user, for this call to fail, rather than to ask the user to check that `proposal.timestamp == 0`. What do you both think, @Orland0x @solimander ?

Orland0x:
> Thanks for your PR! ❤️
> 
> I'm glad you caught these. I'm also glad the code itself is behaving as expected, and it's the "only" the tests that invalid.
> 
> I'll let @Orland0x have a look before we merge.
> 
> One thing that caught my attention is the `reverts when querying an invalid proposal id`. I'm wondering if it would make more sense, for the end user, for this call to fail, rather than to ask the user to check that `proposal.timesetamp == 0`. What do you both think, @Orland0x @solimander ?

Agreed, I think a revert here if the proposal does not exist would be good

solimander:
> One thing that caught my attention is the `reverts when querying an invalid proposal id`. I'm wondering if it would make more sense, for the end user, for this call to fail, rather than to ask the user to check that `proposal.timestamp == 0`. What do you both think, @Orland0x @solimander ?

Agreed and updated 👍

Orland0x:
LGTM

PR #394, opened 2023-01-10
Update devnet

pscott:
While working on #395 I tried updating devnet and dependencies and latest hardhat starknet plugin, which led me to a pretty deep rabbit hole.

Basically:
- Deploying has changed: now it needs to go through accounts
- Those accounts need to be funded, so I added a `getAccount(index)` which attaches to one of the 10 prefunded accounts from the devnet
- I added a `declareAndDeploy` utility function
- For some reason, this broke the nice trick we had to deploy multiple smart contracts in parallel (putting them all as `Promise`s in an array, and calling `await Promise.all(array)`). It simply broke. Idk if came from starknet-devnet or the hardhat plugin or me...
- For some reason I also had to update some tests with a `padLeft` function. Basically starknet addresses are felts which are `251` bits long (at most). Meaning in most cases, an address isn't `64` chars but `63`. However, calling `_signedTypedData` on addresses of odd-length simply doesn't work. So, to help, I added the `padLeft` utility function and used it when necessary.


pscott:
- Linter is broken because I believe `cairo-format` was soft-broken in `0.10.3` (need to double check that).

- Tests are still failing because I have a weird issue with `tx_info.version` where it sometimes shows up as `1` and sometimes as `0x100000000000000000000000000000001`.... need to investigate some more...

pscott:
1) Linter is still broken.
2) Fixed all tests except from the one I mentioned above. Indeed something is odd with the `tx_version`...

I suggest we merge this one, I've [created a tracking issue](https://github.com/snapshot-labs/sx-core/issues/399) for tx_version bug :)

PR #410, opened 2023-05-22
feat: cairo 1 migration

Orland0x:
This PR implements:
- Space state setters and getters.
- Propose, vote, update, execute cancel functions.
- Vanilla versions of each module.
- Eth tx authenticator.
- Eth Sig authenticator.
- L1 execution via an avatar contract.
- A small number of unit tests using the space contract and the vanilla modules


src/authenticators/vanilla.cairo
```diff
@@ -0,0 +1,19 @@
...
+}
+
+#[contract]
+mod VanillaAuthenticator {
+    use starknet::ContractAddress;
+    use starknet::syscalls::call_contract_syscall;
+    use core::array::{ArrayTrait, SpanTrait};
+
+    #[external]
+    fn authenticate(target: ContractAddress, entry_point_selector: felt252, data: Array<felt252>) {
```
arr00: Why not refer to this as `selector` as in ethereum?

src/external/ownable.cairo
```diff
@@ -0,0 +1,54 @@
+// OZ ownable module, adding here directly until OZ releases a Scarb package
```
arr00: No external function to transfer ownership?

src/execution_strategies/vanilla.cairo
```diff
@@ -0,0 +1,22 @@
...
+}
+
+#[contract]
+mod VanillaExecutionStrategy {
+    use starknet::ContractAddress;
+    use sx::utils::types::Proposal;
+
+    struct Storage {
+        _num_executed: felt252
+    }
```
arr00: Is there an autogenerated getter? If not would be useful to confirm executions.

src/utils/bits.cairo
```diff
@@ -0,0 +1,40 @@
...
+
+impl U256BitSetter of BitSetter<u256> {
+    /// Sets the bit at the given index to 1.
+    #[inline(always)]
+    fn set_bit(self: u256, index: u8, bit: bool) -> () {
+        let mask: u128 = pow(2, index.into());
+        let mask: felt252 = mask.into();
+        let mask: u256 = mask.into();
+        if bit {
+            self | mask;
```
arr00: I assume we want to return the expression `self | mask`. I believe the semicolon stops this from happening

src/utils/bits.cairo
```diff
@@ -0,0 +1,40 @@
...
+        // self & mask;
+        }
+    }
+
+    /// Returns true if the bit at the given index is set to 1.
+    #[inline(always)]
+    fn is_bit_set(self: u256, index: u8) -> bool {
+        let mask: u128 = pow(2, index.into());
+        let mask: felt252 = mask.into();
+        let mask: u256 = mask.into();
```
arr00: Confused how putting the mask into a `u128` then converting to `u256` would work

arr00 (commented):
Just some intial comments

src/external/ownable.cairo
Orland0x: i think modules cannot have external functions in cairo1.
 I added external functions `transfer_ownership` and `renounce_ownership`

src/utils/bits.cairo
Orland0x: thanks, actually realised there was a lot wrong there haha. i changed the function to accept a mutable reference which gets updated, so no return value.

src/utils/bits.cairo
Orland0x: yeah ignore that, was some old test code i forgot to remove

pscott (approved):
tACK!

Finally! :D This will of course be an ongoing effort, since Cairo is still evolving and we have quite a few TODOs left :)

Issue #442, opened 2023-07-05
fix: Import U256Zeroable from stdlib


pscott:
https://github.com/snapshot-labs/sx-starknet/blob/687401f87a391c8203d75a261a27ba7639b92f92/starknet/src/utils/math.cairo#L4

U256Zeroable is now part of stdlib, see https://github.com/starkware-libs/cairo/blob/336893775741f10c55fe22d5cd9f7937333df58e/corelib/src/integer.cairo#L1908 .

PR #453, opened 2023-07-07
chore: cairo v2 migration

Orland0x:


starknet/src/authenticators/eth_tx.cairo
```diff
@@ -1,35 +1,34 @@
...
     fn authenticate_update_proposal(
+        ref self: TContractState,
         target: ContractAddress,
         author: ContractAddress,
         proposal_id: u256,
         execution_strategy: Strategy
     );
-    #[l1_handler]
-    fn commit(from_address: felt252, sender_address: felt252, hash: felt252);
+// TODO: Should L1 handlers be part of the interface?
```
pscott, line 28: Hmmm I originally thought "yes" but now I think it shouldn't because users will not be able to call this function anyway so I don't see the point of having it in the interface

starknet/src/execution_strategies/vanilla.cairo
```diff
@@ -1,54 +1,50 @@
...
+    impl VanillaExecutionStrategy of IExecutionStrategy<ContractState> {
         fn execute(
+            ref self: ContractState,
             proposal: Proposal,
             votes_for: u256,
             votes_against: u256,
             votes_abstain: u256,
             payload: Array<felt252>
         ) {
+            // TODO: this is probably wrong.
```
pscott, line 22: 🤣

starknet/src/execution_strategies/vanilla.cairo
```diff
@@ -1,54 +1,50 @@
...
-        votes_for: u256,
-        votes_against: u256,
-        votes_abstain: u256,
-        payload: Array<felt252>
-    ) {
-        VanillaExecutionStrategy::execute(
-            proposal, votes_for, votes_against, votes_abstain, payload
-        );
+    fn constructor(ref self: ContractState, quorum: u256) {
+        // TODO: temporary until components are released
```
pscott, line 40: 🫡

starknet/src/tests/mocks/proposal_validation_always_fail.cairo
```diff
@@ -1,11 +1,20 @@
...
 mod AlwaysFailProposalValidationStrategy {
     use starknet::ContractAddress;
 
-    #[external]
-    fn validate(
-        author: ContractAddress, params: Array<felt252>, userParams: Array<felt252>
-    ) -> bool {
-        false
+    #[storage]
+    struct Storage {}
```
pscott, line 6: Funny quirk that we still have to declare this empty struct

starknet/src/utils/math.cairo
```diff
@@ -1,24 +1,6 @@
 use zeroable::Zeroable;
 use traits::Into;
 
-// TODO: should be able to import this from the standard lib eventually but its not released atm
```
pscott, line 4: Aye! Closes #442

starknet/src/utils/types.cairo
```diff
@@ -115,33 +108,74 @@ impl ProposalStatusIntoU8 of Into<ProposalStatus, u8> {
...
+        lhs.address == rhs.address
+            && poseidon::poseidon_hash_span(
+                lhs.params.span()
+            ) == poseidon::poseidon_hash_span(rhs.params.span())
     }
 
-    fn ne(lhs: FinalizationStatus, rhs: FinalizationStatus) -> bool {
-        FinalizationStatusIntoU8::into(lhs) != rhs.into()
+    fn ne(lhs: @Strategy, rhs: @Strategy) -> bool {
+        !(lhs.clone() == rhs.clone())
```
pscott, line 132: Sucks we need to clone here... I mean in theory we're just checking stuff, not modifying, so I'm guessing there should be a way of not cloning, right?

starknet/src/utils/types.cairo
```diff
@@ -115,33 +108,74 @@ impl ProposalStatusIntoU8 of Into<ProposalStatus, u8> {
...
     }
 }
 
+#[derive(Option, Clone, Drop, Serde)]
+struct IndexedStrategy {
+    index: u8,
+    params: Array<felt252>,
+}
+
+/// NOTE: Using u64 for timestamps instead of u32 which we use in sx-evm. can change if needed.
```
pscott, line 142: What's the rationale for introducing this change?

starknet/src/utils/types.cairo
```diff
@@ -150,360 +184,327 @@ impl StorageAccessFinalizationStatus of StorageAccess<FinalizationStatus> {
...
             FinalizationStatusIntoU8::into(value)
         )
     }
-}
 
-impl LegacyHashChoice of LegacyHash<Choice> {
-    fn hash(state: felt252, value: Choice) -> felt252 {
-        LegacyHash::hash(state, ChoiceIntoU8::into(value))
+    fn size_internal(value: FinalizationStatus) -> u8 {
+        1_u8
```
pscott, line 201: this returns the number of `bytes` used by `value`, right? Just double checking in case it's supposed to be the number of `bits`

starknet/src/utils/types.cairo
```diff
@@ -150,360 +184,327 @@ impl StorageAccessFinalizationStatus of StorageAccess<FinalizationStatus> {
...
             storage_base_address_from_felt252(
                 storage_address_from_base_and_offset(base, 8_u8).into()
             ),
+            offset,
             value.active_voting_strategies
         )
     }
+
+    fn size_internal(value: Proposal) -> u8 {
+        9_u8
```
pscott, line 372: Ah now I see a second implementation, is it supposed to be the number of `word`s ? (`word` being a `felt252`) ?
In that case, couldn't we pack all the `u64` to reduce storage?

pscott (approved):
utACK, just a few questions regarding `u64` and packing :)

starknet/src/utils/types.cairo
Orland0x: no its the number of felts

pscott: yeah sorry, saw that later on here: https://github.com/snapshot-labs/sx-starknet/pull/453#discussion_r1259554803

starknet/src/utils/types.cairo
Orland0x: not sure. the impls in the standard lib dereference the snapshot: https://github.com/starkware-libs/cairo/blob/22ccda9eb66afcb6465e4778ddc310a5d09e3c91/corelib/src/integer.cairo#L297C9-L297C9 so doesnt look like there is a way to compare snapshots directly

starknet/src/utils/types.cairo
Orland0x: its the default type returned by the syscall, not much reasoning atm. We were not packing the timestamps anyway right now so felt unnecessary to cast

starknet/src/utils/types.cairo
Orland0x: yeah we should, i assume this will be handled by the compiler/ another library soon though so not sure its worth writing our own impl

starknet/src/authenticators/eth_tx.cairo
Orland0x: yeah thats currently what the compiler accepts. To me it seems like they should be included though, as its an external entrypoint to the code.

pscott: But only callable by a single address, enforced by the protcol itself so idk ?

starknet/src/utils/types.cairo
pscott: So for `u8` yeah sure but for `param` it seems very expensive to do that.. but yeah maybe it's my `Rust` background talking here... :p 

But like the logic is: if you're comparing values, you def don't need to take ownership of them :p

PR #456, opened 2023-07-13
Refactor settings setters; use snapshots for events

### LINKED ISSUE #427: refactor: use `updateSettings` instead of individual functions
Remove all the `set*` functions and combine them into a single `updateSettings` function.

pscott:
1. Removes all `set*` functions
2. Introduces the `update_settings` function, similarly to sx-evm.
3. Updates events to use `@` (snapshot) rather than values.

Closes #427

### DIFF
--- a/starknet/src/utils/types.cairo
+++ b/starknet/src/utils/types.cairo
@@ -648,8 +648,8 @@ impl UpdateSettingsCalldataImpl of UpdateSettingsCalldataTrait {
             authenticators_to_add: NoUpdateArray::no_update(),
             authenticators_to_remove: NoUpdateArray::no_update(),
             voting_strategies_to_add: NoUpdateArray::no_update(),
-            voting_strategies_metadata_URIs_to_add: array::ArrayTrait::new(),
-            voting_strategies_to_remove: array::ArrayTrait::new(),
+            voting_strategies_metadata_URIs_to_add: NoUpdateArray::no_update(),
+            voting_strategies_to_remove: NoUpdateArray::no_update(),
         }
     }
 }
--- a/starknet/src/utils/types.cairo
+++ b/starknet/src/utils/types.cairo
@@ -7,10 +7,10 @@ use option::OptionTrait;
 use clone::Clone;
 use integer::{U8IntoU128};
 use starknet::{
-    ContractAddress, contract_address_const, StorageAccess, StorageBaseAddress, SyscallResult, storage_write_syscall,
-    storage_read_syscall, storage_address_from_base_and_offset, storage_base_address_from_felt252,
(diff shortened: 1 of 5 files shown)


pscott:
formatting failed but running `cairo-format **/**.cairo` doesn't work because I'm guessing some things have not been updated with v2... have to investigate :p

starknet/src/utils/types.cairo
```diff
@@ -534,3 +535,121 @@ impl IndexedStrategyImpl of IndexedStrategyTrait {
...
+    }
+
+    fn should_update(self: @Array<T>) -> bool {
+        self.len() != 0
+    }
+}
+
+
+impl UpdateSettingsCalldataImpl of UpdateSettingsCalldataTrait {
+    fn default() -> UpdateSettingsCalldata {
```
Orland0x, line 639: super nice to have this for tests

starknet/src/utils/types.cairo
```diff
@@ -534,3 +535,121 @@ impl IndexedStrategyImpl of IndexedStrategyTrait {
...
+            max_voting_duration: NoUpdateU64::no_update(),
+            voting_delay: NoUpdateU64::no_update(),
+            metadata_URI: NoUpdateArray::no_update(),
+            dao_URI: NoUpdateArray::no_update(),
+            proposal_validation_strategy: NoUpdateStrategy::no_update(),
+            proposal_validation_strategy_metadata_URI: NoUpdateArray::no_update(),
+            authenticators_to_add: NoUpdateArray::no_update(),
+            authenticators_to_remove: NoUpdateArray::no_update(),
+            voting_strategies_to_add: NoUpdateArray::no_update(),
+            voting_strategies_metadata_URIs_to_add: array::ArrayTrait::new(),
```
Orland0x: why not use `NoUpdateArray::no_update()`?

Orland0x:
> formatting failed but running `cairo-format **/**.cairo` doesn't work because I'm guessing some things have not been updated with v2... have to investigate :p

might be that `cairo-format` is different to `scarb fmt` we use the latter in the CI. i used scarb fmt and it worked

starknet/src/utils/types.cairo
pscott: Yeah it was a miss; it's because at first I wanted to have something visually different to remind me that `NoUpdateString::no_update()` has not been implemented yet...

We can go with `empty` for now but I think we'll need to change in the future because `""` is a valid metadata_uri and I'd be happier with something like `No Update` as a string to indicate it

Orland0x: so we can change to `NoUpdateArray::no_update()` ?

pscott: aye! will do

PR #458, opened 2023-07-25
chore: upgraded to v2.1.0-rc1

Orland0x:
got a nice LOC reduction from `Store` derives

### DIFF
--- a/starknet/src/utils/types.cairo
+++ b/starknet/src/utils/types.cairo
@@ -5,9 +5,8 @@ use traits::{PartialEq, TryInto, Into};
 use hash::LegacyHash;
 use option::OptionTrait;
 use clone::Clone;
-use integer::{U8IntoU128};
 use starknet::{
-    ContractAddress, contract_address_const, StorageAccess, StorageBaseAddress, SyscallResult,
+    ContractAddress, Store, contract_address_const, StorageBaseAddress, SyscallResult,
     storage_write_syscall, storage_read_syscall, storage_address_from_base_and_offset,
     storage_base_address_from_felt252, contract_address::Felt252TryIntoContractAddress,
     syscalls::deploy_syscall, class_hash::Felt252TryIntoClassHash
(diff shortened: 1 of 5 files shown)


starknet/src/utils/types.cairo
```diff
@@ -37,7 +36,7 @@ enum Choice {
     Abstain: (),
 }
 
-#[derive(Copy, Drop, Serde, PartialEq)]
+#[derive(Drop, Serde, PartialEq, Copy, starknet::Store)]
```
pscott, line 39: why `starknet::Store` and not `Store` ?

starknet/src/utils/types.cairo
```diff
@@ -438,74 +215,8 @@ impl StorageAccessFelt252Array of StorageAccess<Array<felt252>> {
...
-            offset,
-            value.params
-        )
-    }
-
-    fn size_internal(value: Strategy) -> u8 {
-        // Add 1 for the strategy address
-        StorageAccess::size_internal(value.params) + 1
+    fn size() -> u8 {
+        255_u8
```
pscott, line 220: why 255?

starknet/src/utils/types.cairo
Orland0x: for some reason it wouldnt compile with `Store`

Orland0x: not sure why

starknet/src/utils/types.cairo
Orland0x: now size() doesnt take an argument so there is no way to know the size of dynamic types. I spoke to eni about this today and we agreed it made the most sense to put the max size possible. 

Ill add a comment to clarify

pscott: ugh yeah this needs to be take a snapshot... gotcha

starknet/src/utils/types.cairo
pscott: weird... I don't see why it wouldn't but ok !

pscott (approved):
utACK

PR #461, opened 2023-07-25
feat: eth sig auth 

Orland0x:
-


starknet/src/utils/constants.cairo
```diff
@@ -5,40 +5,32 @@ const UPDATE_PROPOSAL_SELECTOR: felt252 =
 
 const ETHEREUM_PREFIX: u128 = 0x1901;
 
-// TODO: Decide on contents of Domain
-// Name, Version, Chain ID, Verifying Contract
-const DOMAIN_TYPEHASH_HIGH: u128 = 0x8b73c3c69bb8fe3d512ecc4cf759cc79;
-const DOMAIN_TYPEHASH_LOW: u128 = 0x239f7b179b0ffacaa9a75d522b39400f;
+// keccak256("EIP712Domain(uint256 chainId)")
```
pscott, line 12: Just the chainId?

starknet/src/utils/eip712.cairo
```diff
@@ -0,0 +1,203 @@
...
+
+// Prefixes a 16 bit prefix to an array of 256 bit values.
+fn _add_prefix_array(input: Array<u256>, mut prefix: u128) -> Array<u256> {
+    let mut out = ArrayTrait::<u256>::new();
+    let mut i = 0_usize;
+    loop {
+        if i >= input.len() {
+            // left shift so that the prefix is in the high bits
+            // let prefix_u256 = u256 { low: 0_128, high: prefix };
+            // let shifted_prefix = prefix_u256 * pow(u256 { low: 2_u128, high: 0_u128 }, 112_u8);
```
pscott: extra

starknet/src/utils/eip712.cairo
```diff
@@ -0,0 +1,203 @@
...
+    let (mut u64_arr, overflow) = into_le_u64_array(encoded_data);
+    keccak::cairo_keccak(ref u64_arr, overflow, 2).byte_reverse()
+}
+
+// Prefixes a 16 bit prefix to an array of 256 bit values.
+fn _add_prefix_array(input: Array<u256>, mut prefix: u128) -> Array<u256> {
+    let mut out = ArrayTrait::<u256>::new();
+    let mut i = 0_usize;
+    loop {
+        if i >= input.len() {
```
pscott: it's better to use `match input.pop_front()` if we can :)

starknet/src/utils/eip712.cairo
```diff
@@ -0,0 +1,203 @@
...
+            // left shift so that the prefix is in the high bits
+            // let prefix_u256 = u256 { low: 0_128, high: prefix };
+            // let shifted_prefix = prefix_u256 * pow(u256 { low: 2_u128, high: 0_u128 }, 112_u8);
+            let shifted_prefix = prefix * pow_u128(2_u128, 112_u8);
```
pscott: need a todo to track when bitshift is added

starknet/src/utils/eip712.cairo
```diff
@@ -0,0 +1,203 @@
...
+        prefix = low_carry;
+        i += 1;
+    };
+    out
+}
+
+// prefixes a 16 bit prefix to a 128 bit input, returning the result and a carry if it overflows 128 bits
+fn _add_prefix_u128(input: u128, prefix: u128) -> (u128, u128) {
+    let prefix_u256 = u256 { low: prefix, high: 0_u128 };
+    let shifted_prefix = prefix_u256 * pow(u256 { low: 2_u128, high: 0_u128 }, 128_u8);
```
pscott: need a todo to track when bitshift is added

starknet/src/utils/endian.cairo
```diff
@@ -0,0 +1,52 @@
+use integer::u128_byte_reverse;
+use sx::utils::math::pow_u128;
+use traits::TryInto;
+use option::OptionTrait;
+use array::ArrayTrait;
+
+const MASK_LOW: u128 = 0xffffffffffffffff; // 2^64 - 1
+const MASK_HIGH: u128 = 0xffffffffffffffff0000000000000000; // 2^128 - 2^64
+const SHIFT: u128 = 0x10000000000000000; // 2^64
```
pscott: never used?

starknet/src/utils/endian.cairo
```diff
@@ -0,0 +1,52 @@
...
+    (
+        low_high.try_into().unwrap(),
+        low_low.try_into().unwrap(),
+        high_high.try_into().unwrap(),
+        high_low.try_into().unwrap()
+    )
+}
+
+// U64 array is little endian
+// Using counter instead of more efficient pop_front approach for looping as we need to keep track of the final index
```
pscott: Not sure I follow this comment.
I see the break condition is `if i == self.len() - 1` ?
I think we could still do something like:

```rust
let mut low_high = 0_u64;
let mut low_low = 0_u64;
let mut high_high = 0_u64;
let mut high_low = 0_u64;

let overflow = loop {
  match self.pop_front() {
    Option::Some(num) => {
      (low_high, low_low, high_high, high_low) = uint256_into_le_words(num);
      if self.len() == 1 {
        break (high_high);
      }
        out.append(high_high);
        out.append(high_low);
        out.append(low_high);
        out.append(low_low);
    },
    Option::None => {
      // panic here would mimic the current implementation
    }
  }
};
```

starknet/src/utils/into.cairo
```diff
@@ -0,0 +1,35 @@
+use array::{ArrayTrait, SpanTrait};
+use option::OptionTrait;
+use traits::Into;
+use integer::u256_from_felt252;
+use starknet::{ContractAddress, EthAddress};
+
+impl Felt252SpanIntoU256Array of Into<Span<felt252>, Array<u256>> {
```
pscott, line 1: I think we can make it generic ?
```rust
impl<T, impl TCopy: Copy<T>, impl TIntoU: Into<T, U>> TSpanIntoU of Into<Span<T>, Array<U>>
  fn into(self: Span<T>) -> Array<U> {
   // ...
  }
```

might need some additional constraints on `U` since it's used by `ArrayTrait`.
If it's too difficult, we can drop the generics on U and just `impl<T> TSpanIntoU256`

starknet/src/utils/into.cairo
```diff
@@ -0,0 +1,35 @@
...
+                Option::None(_) => {
+                    break ();
+                },
+            };
+        };
+        arr
+    }
+}
+
+impl ContractAddressIntoU256 of Into<ContractAddress, u256> {
```
pscott: Same, basically here you are using the fact that self has an `IntoFelt252` impl, so you can make it generic over anything that has a `IntoFelt252` impl I believe :)

```rust
impl<T, impl TIntoFelt252: Into<T, felt252>> TIntoU256 of Into<T, u256>
```

starknet/src/utils/into.cairo
```diff
@@ -0,0 +1,35 @@
...
+    fn into(self: ContractAddress) -> u256 {
+        u256_from_felt252(self.into())
+    }
+}
+
+impl EthAddressIntoU256 of Into<EthAddress, u256> {
```
pscott: with comment frmo above, we could remove this impl

starknet/tests/eth-sig-auth.test.ts
```diff
@@ -0,0 +1,258 @@
...
+  Vote,
+  updateProposalTypes,
+  UpdateProposal,
+} from './types';
+
+const pk = process.env.PRIVATE_KEY || '';
+
+describe('Starknet Signature Authenticator', () => {
+  const signer = new ethers.Wallet(pk);
+  const provider = new Provider({ sequencer: { baseUrl: 'http://127.0.0.1:5050' } });
```
pscott: should this be a constant defined somewhere?

starknet/tests/eth-sig-auth.test.ts
```diff
@@ -0,0 +1,258 @@
...
+  // devnet predeployed account
+  const privateKey0 = '0xe3e70682c2094cac629f6fbed82c07cd';
+  const address0 = '0x7e00d496e324876bbc8531f2d9a82bf154d1a04a50218ee74cdd372f75a551a';
```
pscott: maybe better off as constants so we can use them in other test files ?

starknet/tests/eth-sig-auth.test.ts
```diff
@@ -0,0 +1,258 @@
...
+        _voting_strategies_metadata_URI: [],
+        _authenticators: [ethSigAuthAddress],
+        _metadata_URI: [],
+        _dao_URI: [],
+      }),
+    });
+    spaceAddress = deployResponse.deploy.contract_address;
+
+    domain = {
+      chainId: '0x534e5f474f45524c49', // devnet id
```
pscott: might be nice declaring as const also so we can use in other files

starknet/tests/eth-sig-auth.test.ts
```diff
@@ -0,0 +1,258 @@
...
+      metadataURI: voteMsg.metadataURI,
+    };
+
+    result = await account0.execute({
+      contractAddress: ethSigAuthAddress,
+      entrypoint: 'authenticate_vote',
+      calldata: CallData.compile(voteCalldata as any),
+    });
+
+    console.log(result);
```
pscott: could we assert that the vote was counted correctly? just in case for some reason it didn't revert but didn't count the vote as well

pscott (changes requested):
big PR but looking good!

starknet/tests/eth-sig-auth.test.ts
```diff
@@ -0,0 +1,258 @@
...
+  voteTypes,
+  Vote,
+  updateProposalTypes,
+  UpdateProposal,
+} from './types';
+
+const pk = process.env.PRIVATE_KEY || '';
+
+describe('Starknet Signature Authenticator', () => {
+  const signer = new ethers.Wallet(pk);
```
pscott: why not create a random one ? `new ethers.Wallet.createRandom()`

starknet/tests/eth-sig-auth.test.ts
pscott: btw I think we're better off querying them programatically. They will change at each `starknet-devnet` restart (unless `--seed` is given)

pscott:
now that I think about it, I think we should document how users are expected to run those tests :p

starknet/src/utils/constants.cairo
Orland0x: yeah `name` and `version` need to be encoded as solidity strings which is probably a nightmare to do in cairo. as its not critical to have them in the domain, it didnt feel necessary. 

`verifyingContract` is important though but ether typed data encoder expects an ethereum address. a starknet address cannot be cast to an ethereum one so there is no way for us to represent the authenticator address here. Instead, i added the authenticator to the main message payloads.

starknet/src/utils/eip712.cairo
Orland0x: issue here is that we want to do something different on the last element of the array. so kind of need to track the index instead of just popping each element off. 

using pop front then also tracking an index is probably less efficient than just tracking the index in the loop

Orland0x:
> now that I think about it, I think we should document how users are expected to run those tests :p

yeah its now runnable via `yarn test:eth-sig-auth` so long as they have a devnet running

starknet/src/utils/eip712.cairo
Orland0x: yep moved over

starknet/src/utils/eip712.cairo
Orland0x: yep, i removed all pow calls and replaced with constants anyway

starknet/src/utils/endian.cairo
Orland0x: yeah works

starknet/src/utils/into.cairo
Orland0x: this seems excessive considering we only have 1 impl that would use it

starknet/src/utils/into.cairo
Orland0x: yep agree

starknet/tests/eth-sig-auth.test.ts
Orland0x: dont want to spend to long on make this all super clean. else im just re-inventing starknet hardhat plugin. its mostly just to check the flow works + give wiktor an easy time integrating

starknet/tests/eth-sig-auth.test.ts
Orland0x: ive drawn network url from env var

starknet/tests/eth-sig-auth.test.ts
pscott: gotcha, then I feel like using a `seed` will definitely help for reproducing the tests :)

Orland0x: added a comment with seed 0

Orland0x:
waiting for 2.1.1 to fix the y parity bug before merging.

Orland0x:
> waiting for 2.1.1 to fix the y parity bug before merging.

fixed

pscott:
Would be keen on merging but we should merge https://github.com/snapshot-labs/sx-starknet/pull/513 before and update the constants to make sure tests are running I believe :)

starknet/src/utils/eip712.cairo
```diff
@@ -0,0 +1,201 @@
...
+    hash_typed_data(domain_hash, message_hash)
+}
+
+fn get_domain_hash() -> u256 {
+    // The ethers typed data encoder is not compatible with a Starknet address as the `verifyingContract`
+    // therefore we cannot use the `verifyingContract` field in the domain separator, instead we add the 
+    //  verifying contract address to the message itself.
+    let encoded_data = array![
+        u256 { low: DOMAIN_TYPEHASH_LOW, high: DOMAIN_TYPEHASH_HIGH },
+        Felt252IntoU256::into(get_tx_info().unbox().chain_id)
```
pscott, line 157: `get_tx_info().unbox().chaind_id.into()` would work also I think?

starknet/test/eth-sig-types.ts
```diff
@@ -0,0 +1,94 @@
...
+};
+
+export const proposeTypes = {
+  Propose: [
+    { name: 'authenticator', type: 'uint256' },
+    { name: 'space', type: 'uint256' },
+    { name: 'author', type: 'address' },
+    { name: 'executionStrategy', type: 'Strategy' },
+    { name: 'userProposalValidationParams', type: 'uint256[]' },
+    { name: 'metadataUri', type: 'uint256[]' },
```
pscott: since the `space contract parity` PR got merged, `metadataUri` should be right after `author` (only for `propose`, not for `vote` or `updateProposal`

starknet/test/eth-sig-types.ts
```diff
@@ -0,0 +1,94 @@
...
+  Strategy: sharedTypes.Strategy,
+};
+
+export interface Propose {
+  authenticator: string;
+  space: string;
+  author: string;
+  executionStrategy: Strategy;
+  userProposalValidationParams: string[];
+  metadataUri: string[];
```
pscott: same comment as above

starknet/test/stark-sig-auth.test.ts
```diff
@@ -0,0 +1,256 @@
...
+        _max_voting_duration: 100,
+        _min_voting_duration: 100,
+        _voting_delay: 15,
+        _proposal_validation_strategy: {
+          address: vanillaProposalValidationStrategyAddress,
+          params: [[]],
+        },
+        _proposal_validation_strategy_metadata_URI: [],
+        _voting_strategies: [{ address: vanillaVotingStrategyAddress, params: [] }],
+        _voting_strategies_metadata_URI: [],
```
pscott, line 108: needs to be `[[]]` now that space checks whether `_voting_strategies.len() == _voting_strategies_metadata_uri.len()`

also, all occurences of `URI` should be replaced by `uri` or `Uri` (at least for stuff interacting with cairo contracts)

starknet/test/stark-sig-auth.test.ts
```diff
@@ -0,0 +1,256 @@
...
+      accountType: account0Type,
+    };
+
+    await account0.execute({
+      contractAddress: starkSigAuthAddress,
+      entrypoint: 'authenticate_vote',
+      calldata: CallData.compile(voteCalldata as any),
+    });
+  }, 1000000);
+});
```
pscott, line 256: would be nice to have failing tests by providing wrong signatures / wrong signer

pscott (changes requested):
looking good! added some comments regarding helping debug failing tests :D

starknet/test/stark-sig-auth.test.ts
Orland0x: yeah i have another pr with stark sig tests coming soon

starknet/src/utils/eip712.cairo
Orland0x: Since there is a core lib `Felt252IntoU256` but then we also have the generic `TIntoU256` impl there is 2 impls in scope for the same thing. So need to specify which one to use.

pscott: oh right :) this is one of the quirks of `impl ... of ...` I guess :p (rust doesn't have this issue :p)

pscott (approved):

PR #462, opened 2023-07-25
feat: add factory and upgrade method

pscott:
- Add the `Factory` contract and tests
- Add the `upgrade` function on the `Space` contract (and tests)
- Add the `StrategyImpl` to create Strategies with empty params `StrategyImpl::from_address()`
- Add the `setup` file that returns a `Config` struct that will (hopefully) make testing files leaner

### DIFF
--- a/starknet/src/tests/test_upgrade.cairo
+++ b/starknet/src/tests/test_upgrade.cairo
@@ -65,7 +65,7 @@ mod tests {
 
     #[test]
     #[available_gas(10000000000)]
-    fn test_upgrade_via_exection_strategy() {
+    fn test_upgrade_via_execution_strategy() {
         let config = setup();
         let (factory, space) = deploy(@config);
         let proposal_id = space.next_proposal_id();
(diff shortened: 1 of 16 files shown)


starknet/src/utils/types.cairo
```diff
@@ -120,6 +120,23 @@ struct Strategy {
     params: Array<felt252>,
 }
 
+trait StrategyTrait {
```
Orland0x: do you think this should be defined in the test folder somewhere? as the trait would only ever be used in tests.

starknet/src/tests/test_upgrade.cairo
```diff
@@ -0,0 +1,128 @@
...
+
+        let (execution_contract_address, _) = deploy_syscall(
+            ExecutorExecutionStrategy::TEST_CLASS_HASH.try_into().unwrap(),
+            0,
+            ArrayTrait::<felt252>::new().span(),
+            false
+        )
+            .unwrap();
+
+        // Set the owner to be the space itself
```
Orland0x: Set the owner to be the execution strategy

starknet/src/utils/types.cairo
pscott: bbf7ca6a0345444477f6bbc6adb8e6ea63f5f731

starknet/src/tests/test_upgrade.cairo
pscott: 386c9484ad3e51f6a04354e9c523b10805fd7daa

Orland0x (approved):
LGTM, uTACK

PR #464, opened 2023-07-27
fix: eth tx auth griefing prevention

### LINKED ISSUE #426: EthTx Authenticator Griefing
It's possible for an Ethereum account to grief other accounts that use the `EthTx` authenticator by committing the same hash and overwriting the initial sender address in storage, causing `consume_commit` to revert.

Orland0x:
closes #426


pscott (approved):
utACK

PR #466, opened 2023-08-01
refactor: use abstract UserAddress type for author/voter addresses

Orland0x:
- We switch to a general `UserAddress` enum that can represent both a starknet and ethereum address. this enum can be further extended to include any number of other address types also.
- There is also a custom address type that we represent as a u256. This provides compatibility  for any address type that fits in a u256, without the strongly typed support though.


starknet/src/utils/types.cairo
```diff
@@ -6,14 +6,88 @@ use hash::LegacyHash;
...
+    }
+}
+
+trait UserAddressTrait {
+    fn to_starknet_address(self: UserAddress) -> ContractAddress;
+    fn to_ethereum_address(self: UserAddress) -> EthAddress;
+    fn to_custom_address(self: UserAddress) -> u256;
+}
+
+impl UserAddressImpl of UserAddressTrait {
```
Orland0x: not sure whether theres an easier way to do this

.gitignore
```diff
@@ -3,11 +3,12 @@ node_modules/
...
 
+build/
 cache/
 out/
 
 !/broadcast
 /broadcast/*/31337/
 /broadcast/**/dry-run/
 
-docs/
```
pscott, line 13: sadge

starknet/src/utils/types.cairo
```diff
@@ -6,14 +6,88 @@ use hash::LegacyHash;
...
 
+#[derive(Copy, Drop, Serde, LegacyHash, PartialEq, starknet::Store)]
+enum UserAddress {
+    // Starknet address type
+    StarknetAddress: ContractAddress,
+    // Ethereum address type
+    EthereumAddress: EthAddress,
+    // Custom address type to provide compatibility with any address that can be represented as a u256.
+    CustomAddress: u256
+}
```
pscott: Quick naming comment: the word address is always repeated, seems a bit redundant.

How about:
```
enum Address {
  Starknet: ContractAddress,
  Ethereum: EthAddress,
  Custom: u256,
}
```

starknet/src/utils/types.cairo
```diff
@@ -6,14 +6,88 @@ use hash::LegacyHash;
...
 use array::SpanTrait;
 
+#[derive(Copy, Drop, Serde, LegacyHash, PartialEq, starknet::Store)]
+enum UserAddress {
+    // Starknet address type
+    StarknetAddress: ContractAddress,
+    // Ethereum address type
+    EthereumAddress: EthAddress,
+    // Custom address type to provide compatibility with any address that can be represented as a u256.
+    CustomAddress: u256
```
pscott: Should the customAddress be an `Array::<felt252>` ? something that devs could decode?

starknet/src/utils/types.cairo
```diff
@@ -6,14 +6,88 @@ use hash::LegacyHash;
...
-    ContractAddress, Store, contract_address_const, StorageBaseAddress, SyscallResult,
+    EthAddress, ContractAddress, Store, contract_address_const, StorageBaseAddress, SyscallResult,
     storage_write_syscall, storage_read_syscall, storage_address_from_base_and_offset,
     storage_base_address_from_felt252, contract_address::Felt252TryIntoContractAddress,
     syscalls::deploy_syscall, class_hash::Felt252TryIntoClassHash
 };
 use sx::utils::math::pow;
 use array::SpanTrait;
 
+#[derive(Copy, Drop, Serde, LegacyHash, PartialEq, starknet::Store)]
```
pscott: I believe `LegacyHash` should not be used when possible. Are we sure we wish to derive it?

starknet/src/utils/types.cairo
```diff
@@ -6,14 +6,88 @@ use hash::LegacyHash;
...
+trait UserAddressTrait {
+    fn to_starknet_address(self: UserAddress) -> ContractAddress;
+    fn to_ethereum_address(self: UserAddress) -> EthAddress;
+    fn to_custom_address(self: UserAddress) -> u256;
+}
+
+impl UserAddressImpl of UserAddressTrait {
+    fn to_starknet_address(self: UserAddress) -> ContractAddress {
+        match self {
+            UserAddress::StarknetAddress(address) => address,
```
pscott: in Rust you can do :
```
match self {
  UserAddress::StarknetAddress(address) => address,
  _ => panic_with_felt252('Incorrect address type'),
}
```
does it work in cairo?

pscott (approved):
minor comments but all good!

starknet/src/utils/types.cairo
Orland0x: dynamic type adds overhead so seems unnecessary i think

starknet/src/utils/types.cairo
Orland0x: removed

starknet/src/utils/types.cairo
Orland0x: i dont think so yet. https://book.cairo-lang.org/ch05-02-the-match-control-flow-construct.html#match-0-and-the-_-placeholder

Orland0x: `Unsupported match. Currently, matches require one arm per variant, in the order of variant definition.`

pscott: sucks, weird that the cairo book doesn't mention that. Will probs come in the future

starknet/src/utils/types.cairo
pscott: idea behind it is we don't want to be limited to u256 if a new chain uses bigger addresses but yeah I'm fine with that

PR #467, opened 2023-08-01
feat: stark sig auth

Orland0x:


.eslintrc
```diff
@@ -0,0 +1,22 @@
...
+    "ecmaVersion": 2018,
+    "sourceType": "module"
+  },
+  "rules": {
+    "no-console": "off",
+    "prettier/prettier": "error",
+    "@typescript-eslint/explicit-function-return-type": "off",
+    "@typescript-eslint/ban-ts-ignore": "off",
+    "@typescript-eslint/camelcase": "off",
+    "@typescript-eslint/no-explicit-any": "off"
```
pscott: any reason why these specific ones are off? I think `no-console` could be fine to have, as well as explicit-function-return-type ?

starknet/src/authenticators/stark_sig.cairo
```diff
@@ -0,0 +1,147 @@
...
+    use core::array::{ArrayTrait, SpanTrait};
+    use serde::Serde;
+    use sx::space::space::{ISpaceDispatcher, ISpaceDispatcherTrait};
+    use sx::types::{Strategy, IndexedStrategy, UserAddress, Choice};
+    use sx::utils::stark_eip712;
+
+    #[storage]
+    struct Storage {
+        _domain_hash: felt252,
+        _used_salts: LegacyMap::<(ContractAddress, felt252), bool>
```
pscott, line 54: Is the "standard" to use `_` prefix for storage vars?

pscott, line 54: Is there a Poseidon-based HashMap?

starknet/src/authenticators/stark_sig.cairo
```diff
@@ -0,0 +1,147 @@
...
+                signature,
+                target,
+                author,
+                @execution_strategy,
+                user_proposal_validation_params.span(),
+                salt,
+                account_type
+            );
+
+            self._used_salts.write((author, salt), true);
```
pscott, line 84: Do we need to call something like `.squash()` to make it secure? i don't remember if we needed to do this in cair0

starknet/src/utils/stark_eip712.cairo
```diff
@@ -0,0 +1,167 @@
...
+    }
+};
+use sx::interfaces::{
+    AccountABIDispatcher, AccountABIDispatcherTrait, AccountCamelABIDispatcher,
+    AccountCamelABIDispatcherTrait
+};
+
+fn verify_propose_sig(
+    domain_hash: felt252,
+    signature: Array<felt252>,
```
pscott, line 22: Since we don't consume the signature, would it make more sense to use a `Span<felt252>` ? Same for the other functions

starknet/src/interfaces/i_account.cairo
```diff
@@ -0,0 +1,33 @@
+use array::ArrayTrait;
+use array::SpanTrait;
+use starknet::account::Call;
+use starknet::ContractAddress;
+
+#[starknet::interface]
+trait AccountABI<TState> {
```
pscott, line 8: Where are these taken from? Might be nice to have source link somewhere

starknet/src/utils/constants.cairo
```diff
@@ -42,3 +42,33 @@ const STRATEGY_TYPEHASH_LOW: u128 = 0x78d5506febfdb18580ea361801747e63;
 // keccak256("IndexedStrategy(uint256 index,uint8[] params)")
 const INDEXED_STRATEGY_TYPEHASH_HIGH: u128 = 0x894665428ec742c74109dc21d320d1ab;
 const INDEXED_STRATEGY_TYPEHASH_LOW: u128 = 0x8b36195eec0090e913c01e7534729c74;
+
+
+// ------ Stark Sig Constants ------
+
+// 'StarkNet Message' as short string
+const STARKNET_MESSAGE: felt252 = 0x537461726b4e6574204d657373616765;
```
pscott: I think you can just do
```rust
const STARKNET_MESSAGE: felt252 = 'STARKNET_MESSAGE';
```

starknet/src/utils/constants.cairo
```diff
@@ -42,3 +42,33 @@ const STRATEGY_TYPEHASH_LOW: u128 = 0x78d5506febfdb18580ea361801747e63;
...
+// H('StarkNetDomain(name:felt252,version:felt252,chainId:felt252,verifyingContract:ContractAddress)')
```
pscott, line 52: Which `H` hash function are we using here? I believe it's the Starknet Keccak right? Might be worth specifying

starknet/src/utils/constants.cairo
```diff
@@ -42,3 +42,33 @@ const STRATEGY_TYPEHASH_LOW: u128 = 0x78d5506febfdb18580ea361801747e63;
...
+const STRATEGY_TYPEHASH: felt252 =
+    0x39154ec0efadcd0deffdfc2044cf45dd986d260e59c26d69564b50a18f40f6b;
+// H('IndexedStrategy(index:felt252,params:felt*)')
+const INDEXED_STRATEGY_TYPEHASH: felt252 =
+    0x1f464f3e668281a899c5f3fc74a009ccd1df05fd0b9331b0460dc3f8054f64c;
+// H('u256(low:felt252,high:felt252)')
+const U256_TYPEHASH: felt252 = 0x1094260a770342332e6a73e9256b901d484a438925316205b4b6ff25df4a97a;
+
+// ------ ERC165 Interface Ids ------
+const ERC165_ACCOUNT_INTERFACE_ID: felt252 = 0xa66bd575; // snake
```
pscott, line 82: Might be worth specifying how they're obtained too

starknet/src/utils/signatures.cairo
```diff
@@ -72,6 +72,7 @@ impl KeccakTypeHashIndexedStrategyArray of KeccakTypeHash<Array<IndexedStrategy>
                 break ();
             }
             encoded_data.append(self.at(i).clone().hash());
+            i += 1;
```
pscott, line 75: Oupsy

pscott: Btw now that I think about it, I think it should be
```
trait KeccakTypeHash<T> {
    fn hash(self: @T) -> u256;
}
```

don't you think? Doesn't make sense to consume the `T` here (can create an issue and change this in another PR)

starknet/src/utils/stark_eip712.cairo
```diff
@@ -0,0 +1,167 @@
...
+    message_hash.serialize(ref encoded_data);
+    encoded_data.span().struct_hash()
+}
+
+/// Verifies the signature of a message by calling the account contract.
+fn verify_signature(
+    digest: felt252, signature: Array<felt252>, account: ContractAddress, account_type: felt252
+) {
+    if account_type == 'snake' {
+        assert(
```
pscott, line 164: How easy is it to know the `account_type` for the front end? Because we could also imagine not having `account_type` and calling `supportsInterace(ERC165_ACCOUNT_INTERFACE_ID)`, checking if it's true, if not call the `OLD_ACCOUNT`, check if it's valid. Adds a little bit more loads on the smart contract (but not that much because most accounts will support `ERC165_ACCOUNT_INTERFACE_ID`, so in most cases the gas cost will be similar to the current implementation

starknet/src/utils/stark_eip712.cairo
```diff
@@ -0,0 +1,167 @@
...
+    } else if account_type == 'camel' {
+        assert(
+            AccountABIDispatcher {
+                contract_address: account
+            }.supports_interface(ERC165_OLD_ACCOUNT_INTERFACE_ID) == true,
+            'Invalid Account'
+        );
+        AccountABIDispatcher { contract_address: account }.is_valid_signature(digest, signature);
+    } else {
+        panic_with_felt252('Invalid Account Type');
```
pscott, line 180: We could use an enum for the `account_type` this way we don't have this panic! :)

starknet/src/utils/struct_hash.cairo
```diff
@@ -0,0 +1,69 @@
...
+    fn struct_hash(self: @IndexedStrategy) -> felt252 {
+        let mut encoded_data = ArrayTrait::<felt252>::new();
+        INDEXED_STRATEGY_TYPEHASH.serialize(ref encoded_data);
+        (*self.index).serialize(ref encoded_data);
+        self.params.span().struct_hash().serialize(ref encoded_data);
+        encoded_data.span().struct_hash()
+    }
+}
+
+impl StructHashIndexedStrategySpan of StructHash<Span<IndexedStrategy>> {
```
pscott, line 45: instead of `Span<felt252>` we could:
```rust
impl StructHashSpan<T, impl TStructHash: StructHash<T>> of StructHash<Span<T>> {
    fn struct_hash(self: @Span<IndexedStrategy>) -> felt252 {
        let mut self_ = *self;
        let mut encoded_data = ArrayTrait::<felt252>::new();
        loop {
            match self_.pop_front() {
                Option::Some(item) => {
                    encoded_data.append(item.struct_hash());
                },
                Option::None(_) => {
                    break encoded_data.span().struct_hash();
                },
            };
        }
    }
}
```

and if you implement `StructHash<felt252>` simply as
```rust
#[inline(always)]
struct_hash(self: @felt252) -> felt252 {
  *self
}
```

then you can remove this impl and the `impl StructHashSpanFelt252` also :D

starknet/src/utils/struct_hash.cairo
```diff
@@ -0,0 +1,69 @@
...
+                },
+                Option::None(_) => {
+                    break encoded_data.span().struct_hash();
+                },
+            };
+        }
+    }
+}
+
+impl StructHashU256 of StructHash<u256> {
```
pscott, line 62: also you could define a trait
```rust
trait TypeHash<T> {
  fn type_hash() -> felt252;
}
```


```
impl StructHashSerde<T, impl TSerde: Serde<T>, impl TTypeHash: TypeHash<T>> of StructHash<T> {
  fn struct_hash(self: @T) -> felt252 {
    let mut encoded_data = ArrayTrait::<felt252>::new();
    TypeHash::<T>::type_hash().serialize(ref encoded_data);
    self.serialize(ref encoded_data);
    encoded_data.span().struct_hash()
}
```

and then you can remove all those impls :D

GENERICS BABY

starknet/tests/stark-sig.test.ts
```diff
@@ -0,0 +1,230 @@
...
+  Propose,
+  Vote,
+  UpdateProposal,
+  StarknetSigProposeCalldata,
+  StarknetSigVoteCalldata,
+  StarknetSigUpdateProposalCalldata,
+} from './types';
+
+describe('Starknet Signature Authenticator', () => {
+  const provider = new Provider({ sequencer: { baseUrl: 'http://127.0.0.1:5050' } });
```
pscott: maybe we should define this as a const somewhere?

starknet/tests/stark-sig.test.ts
```diff
@@ -0,0 +1,230 @@
...
+      fs.readFileSync('starknet/target/dev/sx_Space.casm.json').toString('ascii'),
+    );
+
+    deployResponse = await account0.declareAndDeploy({
+      contract: spaceSierra,
+      casm: spaceCasm,
+      constructorCalldata: CallData.compile({
+        _owner: 1,
+        _max_voting_duration: 100,
+        _min_voting_duration: 100,
```
pscott, line 101: i see that the constructor is `max / min` but idk why my brain prefers `min / max` order. wdyt?

starknet/tests/stark-sig.test.ts
```diff
@@ -0,0 +1,230 @@
...
+    };
+
+    const proposeData: typedData.TypedData = {
+      types: proposeTypes,
+      primaryType: 'Propose',
+      domain: domain,
+      message: proposeMsg as any,
+    };
+
+    const proposeSig = (await account0.signMessage(proposeData)) as any;
```
pscott, line 145: love the ts `as any` lol :p should we have a `Signature` type?

pscott (changes requested):
looks nice, most comments are about style :)

starknet/src/utils/signatures.cairo
Orland0x: yeah ignore the eth sig stuff ive changed a lot of that and will submit a separate PR this week

starknet/src/utils/stark_eip712.cairo
Orland0x: the issue is that theres not a standard `supports_interface` interface! (Some use `supportsInterface`.) Also theres no try catch logic yet in cario 1, so we cannot try one interface then the other.

starknet/src/utils/struct_hash.cairo
Orland0x: ooo very nice. saving millions of $ in audit costs in the process lol

starknet/src/authenticators/stark_sig.cairo
Orland0x: no not needed

starknet/src/authenticators/stark_sig.cairo
Orland0x: ive only seen `LegacyHash` be used. I think thats enforced at by the protocol atm

Orland0x: we could not, but then we would get a name collision with getters (not in this example but would in the space) - so would need to rename getters (`get_...`) etc.  

idm the `_`, makes it immediately clear you are reading a storage var when you see one. 

https://github.com/OpenZeppelin/cairo-contracts/blob/495ed8a66cc6f3e6ca7eb8cb741ff4eba2265807/src/access/ownable/ownable.cairo#L53C3-L53C3

pscott: k... I see that `LegacyHash` is deprecated... https://twitter.com/gaetbout/status/1688442688457084928?s=20

guess it'll take time before it's actually deprecated

Orland0x: ah we should migrate all usage then. created an issue to do separately #486

starknet/src/utils/stark_eip712.cairo
Orland0x: the account interface expects an `Array<felt252>` though.

starknet/src/interfaces/i_account.cairo
Orland0x: from OZ, will add comment

starknet/src/utils/constants.cairo
Orland0x: fixed

starknet/src/utils/stark_eip712.cairo
Orland0x: excessive complexity maybe?

starknet/src/utils/struct_hash.cairo
Orland0x: in discord dms decided against this approach

starknet/src/utils/struct_hash.cairo
Orland0x: not convinced this is worth it. adds significant overhead to `StructHashSpanFelt252` as you are essentially just copying the array for no reason. Would make sense if we had other arrays of structs i agree

starknet/tests/stark-sig.test.ts
Orland0x: yeah can reorder, doesnt rly make sense to do max min

starknet/tests/stark-sig.test.ts
Orland0x: added as env var

starknet/tests/stark-sig.test.ts
Orland0x: created issue #487

starknet/tests/stark-sig.test.ts
Orland0x: not gunna bother typing everything for the tests. im sure wiktor will do something nice for sx.js ;)

starknet/src/utils/constants.cairo
Orland0x: added link to the snip

starknet/src/utils/constants.cairo
Orland0x: yep

.eslintrc
Orland0x: i just copied from cairo 0 version of the repo. agree though, will update

Orland0x: went for recommended lints only with the addition of allowing explicit `any` as strong typing ts seems excessive for tests + prettier

starknet/src/authenticators/stark_sig.cairo
pscott: Ok, then I think we should make it a standard accross all contracts if possible

https://github.com/snapshot-labs/sx-starknet/issues/489

starknet/src/utils/stark_eip712.cairo
pscott: It's an `#[external]` interface right? Are you sure you can't pass a Span to an external `Array` ? Find it odd lol

starknet/src/utils/stark_eip712.cairo
pscott: I actually think it'd be better because then it would be enforced when the tx is created / cast.. like it would instantly fail instead of going through all the code and failing here!

starknet/src/utils/stark_eip712.cairo
pscott: yeah got it. sucks but that's how it is!

pscott (approved):
utACK

starknet/src/utils/stark_eip712.cairo
Orland0x: no you cant i dont think - maybe you can modify the interface though as it shouldnt change the function signature

PR #469, opened 2023-08-03
feat: merkle whitelist

### LINKED ISSUE #468: feat: Merkle Whitelist
Whitelist voting strategy with weights.
Must accept both potentially for both L1 (EVM) and L2 (starknet) addresses.

pscott:
Reimplementation of the previous merkle whitelist (https://github.com/snapshot-labs/sx-starknet/blob/f81bab3a1de328d8ee657b8a47678f8dccf4b468/contracts/starknet/VotingStrategies/MerkleWhitelist.cairo)

The main difference here is that the address is a `UserAddress` (2 felt 252) and not a `felt252` anymore.

Also comes with utils for proof creation in cairo (in `sx::test_merkle_whitelist::merkle_utils`)

Closes #468

### DIFF
--- a/starknet/src/utils/merkle.cairo
+++ b/starknet/src/utils/merkle.cairo
@@ -2,14 +2,14 @@ use core::traits::Into;
 use array::{ArrayTrait, Span, SpanTrait};
 use option::OptionTrait;
 use serde::Serde;
-use starknet::ContractAddress;
+use sx::types::UserAddress;
 use clone::Clone;
 use hash::{LegacyHash};
 use debug::PrintTrait;
 
 #[derive(Copy, Clone, Drop, Serde)]
 struct Leaf {
-    address: ContractAddress, // use UserAddress
+    address: UserAddress,
     voting_power: u256,
 }
 
--- a/starknet/src/utils/merkle.cairo
+++ b/starknet/src/utils/merkle.cairo
@@ -6,6 +6,7 @@ use sx::types::UserAddress;
 use clone::Clone;
 use hash::{LegacyHash};
 use debug::PrintTrait;
+use sx::utils::legacy_hash::LegacyHashSpan;
 
 /// Leaf struct for the merkle tree
 #[derive(Copy, Clone, Drop, Serde)]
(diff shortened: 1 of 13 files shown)


starknet/src/utils/merkle.cairo
```diff
@@ -0,0 +1,74 @@
...
+struct Leaf {
+    address: UserAddress,
+    voting_power: u256,
+}
+
+trait Hash<T> {
+    fn hash(self: @T) -> felt252;
+}
+
+impl LegacyHashSpan of LegacyHash<Span<felt252>> {
```
Orland0x: also add this impl in stark sig auth pr so can remove from whichever pr we merge second.

starknet/src/utils/merkle.cairo
```diff
@@ -0,0 +1,74 @@
...
+    assert(computed_root == root, 'Merkle: Invalid proof');
+}
+
+/// Internal helper function that computes the merkle root, given a leaf node and a proof.
+fn _compute_merkle_root(mut current: felt252, proof: Span<felt252>) -> felt252 {
+    let mut proof = proof;
+    loop {
+        match proof.pop_front() {
+            Option::Some(val) => {
+                let p_u256: u256 = (*val).into(); // Needed for type annotation
```
Orland0x, line 44: why do we need to convert to `u256` for the comparison?

starknet/src/voting_strategies/merkle_whitelist.cairo
```diff
@@ -0,0 +1,38 @@
...
+    #[external(v0)]
+    impl MerkleWhitelistImpl of IVotingStrategy<ContractState> {
+        fn get_voting_power(
+            self: @ContractState,
+            block_number: u32,
+            voter: UserAddress,
+            params: Array<felt252>, // [root]
+            user_params: Array<felt252>, // [Serde(leaf), Serde(proofs)]
+        ) -> u256 {
+            let LEAF_SIZE = 4; // Serde::<Leaf>::serialize().len()
```
Orland0x: could define this as a `const` at top of file or in `constants.cairo`

starknet/src/utils/merkle.cairo
pscott: felt252 doesn't impl PartialEq

starknet/src/utils/merkle.cairo
pscott: Idk how we should organize files for this... maybe put it in its own file for easy import?

Orland0x: organizing by trait probably makes the most sense. ive got a `legacy_hash.cairo` file with all the legacy hash impls (currently `Span<felt252>` and `EthAddress`)

starknet/src/voting_strategies/merkle_whitelist.cairo
pscott: went for top of the file because I think it's cleaner as I'm expecting it's use to stay somewhat local

Orland0x (approved):
uTACK

PR #470, opened 2023-08-03
refactor: time tracking

Orland0x:
To match the sx-evm architecture post audit, we make the following changes to time tracking: 
 - migrates u64 timestamps to u32 block numbers for `voting_delay`, `min_voting_duration` and `max_voting_duration`. 
- Moves the snapshot timestamp to be the start of the voting period.

### DIFF
--- a/starknet/src/interfaces/i_voting_strategy.cairo
+++ b/starknet/src/interfaces/i_voting_strategy.cairo
@@ -5,7 +5,7 @@ use sx::types::UserAddress;
 trait IVotingStrategy<TContractState> {
     fn get_voting_power(
         self: @TContractState,
-        timestamp: u64,
+        timestamp: u32,
         voter: UserAddress,
         params: Array<felt252>,
         user_params: Array<felt252>,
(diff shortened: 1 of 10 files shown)


starknet/src/space/space.cairo
```diff
@@ -288,7 +289,10 @@ mod Space {
             let mut proposal = self._proposals.read(proposal_id);
             assert_proposal_exists(@proposal);
             assert(proposal.author == author, 'Only Author');
-            assert(info::get_block_timestamp() < proposal.start_timestamp, 'Voting period started');
+            assert(
+                info::get_block_number().try_into().unwrap() < proposal.start_block_number,
```
pscott: I think we're better off not using `.try_into().unwrap()` here and doing
`info::get_block_number < proposal.start_block_number.into()` ? will avoid having some match and panics

starknet/src/interfaces/i_voting_strategy.cairo
```diff
@@ -5,7 +5,7 @@ use sx::types::UserAddress;
 trait IVotingStrategy<TContractState> {
     fn get_voting_power(
         self: @TContractState,
-        timestamp: u64,
+        timestamp: u32,
```
pscott: should this be renamed to `block_number`?

pscott: same for the voting strategies

pscott (approved):
utACK

starknet/src/space/space.cairo
Orland0x: yep changed

PR #475, opened 2023-08-11
feat: metadata URIs

Orland0x:
Adds missing metadata URIs: space, dao, vote,  propose


pscott (approved):
utACK:

edit: re-review

pscott (changes requested):
I think we're missing some changes in `constants.cairo` https://github.com/snapshot-labs/sx-starknet/blob/dd93f6a50fc28370043fb1dfb9a5c05758e3063f/starknet/src/utils/constants.cairo#L24

the SX-EVM version have the metadataURI in the type hash:
https://github.com/snapshot-labs/sx-evm/blob/df37a70673067300a3a6c0541740117a86d0bc59/src/utils/SignatureVerifier.sol#L33

pscott (approved):
utack

PR #477, opened 2023-08-12
feat: stark tx auth

Orland0x:


starknet/src/authenticators/stark_tx.cairo
```diff
@@ -0,0 +1,108 @@
...
+    impl StarkTxAuthenticator of IStarkTxAuthenticator<ContractState> {
+        fn authenticate_propose(
+            ref self: ContractState,
+            space: ContractAddress,
+            author: ContractAddress,
+            execution_strategy: Strategy,
+            user_proposal_validation_params: Array<felt252>,
+            metadata_URI: Array<felt252>
+        ) {
+            assert(info::get_caller_address() == author, 'Invalid Caller');
```
pscott, line 54: should we assert `author != 0` ? i remember `get_caller_address` could return `0` in the past version of starknet, and I feel like it would add some safety without costing too much?

starknet/src/authenticators/stark_tx.cairo
```diff
@@ -0,0 +1,108 @@
...
+        fn authenticate_vote(
+            ref self: ContractState,
+            space: ContractAddress,
+            voter: ContractAddress,
+            proposal_id: u256,
+            choice: Choice,
+            user_voting_strategies: Array<IndexedStrategy>,
+            metadata_URI: Array<felt252>
+        ) {
+            assert(info::get_caller_address() == voter, 'Invalid Caller');
```
pscott, line 76: same comment as above

starknet/src/authenticators/stark_tx.cairo
pscott: I see we didn't enforce it in SX-EVM and I know the proposal validation strategy will probably return `0` for null address but... I feel like this is the sort of subtleties that might introduce a bug later on, if say we use a weird proposal validation strategy that actually returns something valid when author is 0 ?

starknet/src/authenticators/stark_tx.cairo
```diff
@@ -0,0 +1,108 @@
...
+
+        fn authenticate_update_proposal(
+            ref self: ContractState,
+            space: ContractAddress,
+            author: ContractAddress,
+            proposal_id: u256,
+            execution_strategy: Strategy,
+            metadata_URI: Array<felt252>
+        ) {
+            assert(info::get_caller_address() == author, 'Invalid Caller');
```
pscott, line 98: same as above

starknet/src/tests/test_stark_tx_auth.cairo
```diff
@@ -0,0 +1,313 @@
...
+        UpdateSettingsCalldataImpl, UpdateSettingsCalldata
+    };
+    use sx::tests::utils::strategy_trait::{StrategyImpl};
+
+    fn setup_stark_tx_auth(
+        space: ISpaceDispatcher, owner: ContractAddress
+    ) -> IStarkTxAuthenticatorDispatcher {
+        // Deploy Stark Tx Authenticator and whitelist in Space
+        let (stark_tx_authenticator_address, _) = deploy_syscall(
+            StarkTxAuthenticator::TEST_CLASS_HASH.try_into().unwrap(), 0, array![].span(), false
```
pscott, line 32: `array![]` ?! ?! the array macro is here? how did I not know that ?!

starknet/src/tests/test_stark_tx_auth.cairo
```diff
@@ -0,0 +1,313 @@
...
+        updateSettingsCalldata.authenticators_to_add = array![stark_tx_authenticator_address];
+        testing::set_contract_address(owner);
+        space.update_settings(updateSettingsCalldata);
+
+        authenticator
+    }
+
+    #[test]
+    #[available_gas(10000000000)]
+    fn test__propose_update_vote() {
```
pscott: why two underscores?

starknet/src/tests/test_stark_tx_auth.cairo
```diff
@@ -0,0 +1,313 @@
...
+        let config = setup();
+        let (factory, space) = deploy(@config);
+        let authenticator = setup_stark_tx_auth(space, config.owner);
+
+        let quorum = u256_from_felt252(1);
```
pscott, line 55: `1_u256`

pscott: actually can grep and replace `u256_from_felt252(constant)`

starknet/src/tests/test_stark_tx_auth.cairo
```diff
@@ -0,0 +1,313 @@
...
+        let mut constructor_calldata = ArrayTrait::<felt252>::new();
```
pscott, line 56: can we use `let mut constructor_calldata: Array<felt252> = array![]` here?

starknet/src/tests/test_stark_tx_auth.cairo
```diff
@@ -0,0 +1,313 @@
...
+        // Create Proposal
+        testing::set_contract_address(author);
+        authenticator
+            .authenticate_propose(
+                space.contract_address,
+                author,
+                vanilla_execution_strategy,
+                ArrayTrait::<felt252>::new(),
+                ArrayTrait::<felt252>::new()
+            );
```
pscott, line 203: `array![]` ?

pscott (approved):
utACK

starknet/src/tests/test_stark_tx_auth.cairo
Orland0x: yes its here! should migrate a lot of other stuff to it too while we can

starknet/src/tests/test_stark_tx_auth.cairo
Orland0x: was the syntax i was using for tests. not necessary though, will remove

starknet/src/tests/test_stark_tx_auth.cairo
Orland0x: oh theres a lot of other places where i used `u256_from_felt252` or even u256 { 1, 0} so will fix on a diff pr

starknet/src/tests/test_stark_tx_auth.cairo
pscott: https://github.com/snapshot-labs/sx-starknet/issues/481

starknet/src/tests/test_stark_tx_auth.cairo
Orland0x: yep

starknet/src/tests/test_stark_tx_auth.cairo
Orland0x: yep

starknet/src/authenticators/stark_tx.cairo
Orland0x: agreed, shouldnt it be enforced in the space though?

starknet/src/tests/test_stark_tx_auth.cairo
Orland0x: im not sure how to do this while serializing the types into the array of felts. any ideas?

pscott: I don't think you can do this in a one liner but here I was just referring to using `array![]` instead of `ArrayTrait::<felt252>::new()`

pscott: serializing will still need to be done on the line just after i'm afraid

starknet/src/authenticators/stark_tx.cairo
pscott: probably yeah, this would lower the burden on authenticators.  https://github.com/snapshot-labs/sx-starknet/issues/488

PR #482, opened 2023-08-15
fix: throw if commit is not found

Sekhmet:
Regression from: https://github.com/snapshot-labs/sx-starknet/blame/c5a728759445c56276eeb1e7423bdc0664ffe86e/starknet/src/authenticators/eth_tx.cairo#L131


pscott (approved):
utack

PR #492, opened 2023-08-16
feat: prevent zero address for author/voter

### LINKED ISSUE #488: Check 0 address in space contract
https://github.com/snapshot-labs/sx-starknet/pull/477#discussion_r1294924581

Orland0x:
closes #488


starknet/src/tests/test_space.cairo
```diff
@@ -333,4 +333,115 @@ mod tests {
...
+        let author = UserAddress::Custom(0_u256);
+        let mut propose_calldata = array::ArrayTrait::<felt252>::new();
+        author.serialize(ref propose_calldata);
+        vanilla_execution_strategy.serialize(ref propose_calldata);
+        ArrayTrait::<felt252>::new().serialize(ref propose_calldata);
+        ArrayTrait::<felt252>::new().serialize(ref propose_calldata);
+
+        // Create Proposal
+        authenticator.authenticate(space.contract_address, PROPOSE_SELECTOR, propose_calldata);
+    }
```
pscott, line 571: Erm I know you wrote a lot of the code but I would suggest to do these tests:

write a test module that only does a unit test on `is_zero` (could actually be inside of `user_address.cairo` tbh but doesn't have to).
In this unit test we try starknet, ethereum, and custom).

And then we actually test the three entrypoints (to get full code coverage and regression test), on `cancel`, `propose`, `update_proposal`, and `vote`. This time we don't care about whether they're starknet or ethereum, we just check to see if the code is here :)

starknet/src/space/space.cairo
```diff
@@ -191,6 +191,7 @@ mod Space {
             metadata_URI: Array<felt252>,
```
pscott, line 256: any reason we're not checking this for `cancel`?

If a DAO renounces ownership (sets owner to `0`), and then for some reason in a later starknet update there's a way to create txs with address zero (like when you deploy a contract or something...) then we'd be fucked. Doesn't cost much to add a check for `cancel` too imho! :)

pscott (changes requested):
sorry I forgot to click "submit review" yesterday :(

starknet/src/space/space.cairo
Orland0x: wdym? like preventing renounce ownership?

pscott: hmm sorry i wasn't clear.
my question was " why no check on the `fn cancel()` function ?

Orland0x: checking the caller address is not zero?

pscott: yes

Orland0x: seems like a crazy situation, would break so much if so that was the case. I think its safe to assume a caller cannot be the zero address. We would also need to check all setters etc too

Reason why its worth checking for `author`/`voter` is that its an argument so could quite easily be set to zero accidentally

pscott: Yeah I guess you're right, it would break a lots of other stuff, not worth the trouble :)

starknet/src/tests/test_space.cairo
Orland0x: yep got the unit tests and all entry points tested. sorry was being a bit lazy

pscott: nice, perfect

pscott (approved):
utACK, added some false positive tests

PR #499, opened 2023-08-17
chore: use u256 const syntax

### LINKED ISSUE #480: refactor: u256 const syntx
use `c_u256` for u256 consts instead of more verbose syntax.

pscott:
Closes #480


Orland0x (approved):
uTACK

PR #503, opened 2023-08-17
feat: struct packing

### LINKED ISSUE #483: refactor: struct packing
There is no packing on structs atm. this mostly affects the cost of creating a proposal, as its uses a lot more storage slots than necessary. Long term i assume this will be handled by the compiler within the `Store` derive, but this in unlikely to exist in the next few weeks.

pscott:
Implement struct packing (for Proposal, as the Storage struct will not be modified as much) using the [`StorePacking` trait](https://github.com/starkware-libs/cairo/blob/bc7ca90225e114c43715cd11415e5d9100b5d2c6/corelib/src/starknet/storage_access.cairo#L68)

Closes #483


starknet/src/types/proposal.cairo
```diff
@@ -1,16 +1,231 @@
...
-    finalization_status: FinalizationStatus,
     active_voting_strategies: u256
 }
+
+#[derive(Drop, starknet::Store)]
+struct PackedProposal {
+    timestamps_and_finalization_status: u128, // In order: start, min, max, finalization_status
+    execution_payload_hash: felt252,
+    execution_strategy: ContractAddress,
+    author: UserAddress,
```
Orland0x, line 43: Annoying that the `UserAddress` adds storage overhead. one solution could be to store a `felt252` hash of it that we then compute when neeeded. probably unnecessary complexity though

starknet/src/types/proposal.cairo
```diff
@@ -1,16 +1,231 @@
...
+    active_voting_strategies: u256,
+}
+
+impl ProposalStorePacking of StorePacking<Proposal, PackedProposal> {
```
Orland0x, line 47: really like cairo's approach to packing. super clean and efficient

Orland0x (approved):
uTACK

starknet/src/types/proposal.cairo
pscott: Technically, most of the time the `UserAddress` will be ethereum / starknet and will felt in a single felt so the storage will not move (will stay 0) so i don't think it will cost that much?

starknet/src/types/proposal.cairo
pscott: yeah! agreed

pscott: one thing is make sure you `#[derive(starknet::Store)]` and not `#[derive(Store)]`. Found out the hard way...

PR #505, opened 2023-08-18
Fix missing checks in execute and use `match` in loops

### LINKED ISSUE #501: Missing execution payload hash check and finalization status check
https://github.com/snapshot-labs/sx-starknet/blob/7128eef8e2518955c764539314a7aec2447357ee/starknet/src/space/space.cairo#L285

Missing two checks, check with Sx-EVM version: https://github.com/snapshot-labs/sx-evm/blob/main/src/Space.sol#L280

pscott:
Fixes some missing checks in execute, update proposal, and min/max setters.
Also uses `match` in loops when possible.

Closes #501


Orland0x:
we also need to update `_get_cumulative_power` in the space, still uses inefficient looping.

pscott:
> we also need to update `_get_cumulative_power` in the space, still uses inefficient looping.

yeah I thought I'd do it in another PR but you're right. Updated `_get_cumulative_power` and its function signature to use `Span` rather than `Array`

Orland0x (approved):
uTACK

PR #506, opened 2023-08-19
Add update settings tests

pscott:


starknet/src/tests/test_update_settings.cairo
```diff
@@ -0,0 +1,279 @@
...
+        assert(
+            space.min_voting_duration() == input.min_voting_duration,
+            'Min voting duration not updated'
+        );
+    // TODO: check event once it's been added
+    }
+
+    #[test]
+    #[available_gas(10000000000)]
+    // #[should_panic(expected: ('Invalid duration', 'ENTRYPOINT_FAILED'))] // TODO: uncomment once PR is merged
```
Orland0x: which PR is this?

pscott: #505 fixed the `update_settings` function

Orland0x:
We need to actually store the daoURI in the state in order to be compatible with the standard like we do in sx-evm. I forgot to do this before
https://github.com/snapshot-labs/sx-evm/blob/df37a70673067300a3a6c0541740117a86d0bc59/src/Space.sol#L49C12-L49C12

pscott:
> We need to actually store the daoURI in the state in order to be compatible with the standard like we do in sx-evm. I forgot to do this before https://github.com/snapshot-labs/sx-evm/blob/df37a70673067300a3a6c0541740117a86d0bc59/src/Space.sol#L49C12-L49C12

Yup, funny you mention it, I'm on it with [this PR](https://github.com/snapshot-labs/sx-starknet/pull/529), maybe we can merge this one now so I re-update the `update_settings` function?

Orland0x:
> > We need to actually store the daoURI in the state in order to be compatible with the standard like we do in sx-evm. I forgot to do this before https://github.com/snapshot-labs/sx-evm/blob/df37a70673067300a3a6c0541740117a86d0bc59/src/Space.sol#L49C12-L49C12
> 
> Yup, funny you mention it, I'm on it with [this PR](https://github.com/snapshot-labs/sx-starknet/pull/529), maybe we can merge this one now so I re-update the `update_settings` function?

yup cool

Orland0x (approved):
uTACK

PR #511, opened 2023-08-21
Fix erc20 proposition strategy

pscott:
- Use `timestamp - 1` for proposing power proposal validation strategy
- Add test to ensure it's working correctly
- Fix naming in `get_voting_power` (used `block_number` instead of `timestamp`)


Orland0x (approved):
uTACK

PR #513, opened 2023-08-21
lowercase acronyms

### LINKED ISSUE #508: lowercase acronyms

pscott:
Closes #508 

Needs to get merged before #461 and #461 needs to update the constants (here I updated the comments but did not update the values because there are no tests anyway)


Orland0x (approved):
uTACK

PR #517, opened 2023-08-24
fix:  event syntax

Orland0x:


starknet/src/factory/factory.cairo
```diff
@@ -24,15 +24,24 @@ mod Factory {
...
+    }
 
     #[storage]
     struct Storage {}
 
     #[external(v0)]
     impl Factory of IFactory<ContractState> {
         fn deploy(
-            self: @ContractState,
+            ref self: ContractState,
```
Orland0x, line 44: method needed to have write permissions in order to emit an event.

starknet/src/space/space.cairo
```diff
@@ -490,63 +582,126 @@ mod Space {
...
-                    @input.proposal_validation_strategy,
-                    @input.proposal_validation_strategy_metadata_URI
-                );
+                self
+                    .emit(
+                        Event::ProposalValidationStrategyUpdated(
+                            ProposalValidationStrategyUpdated {
+                                proposal_validation_strategy: input
+                                    .proposal_validation_strategy
+                                    .clone(),
```
Orland0x, line 644: not sure whether there is a way to avoid the clones of structs in the events.  

Now events take a struct, you cannot pass a snapshot to one of the fields like we did before.

starknet/src/space/space.cairo
```diff
@@ -118,88 +119,135 @@ mod Space {
...
+        VoteCast: VoteCast,
+        ProposalExecuted: ProposalExecuted,
+        ProposalUpdated: ProposalUpdated,
+        ProposalCancelled: ProposalCancelled,
+        VotingStrategiesAdded: VotingStrategiesAdded,
+        VotingStrategiesRemoved: VotingStrategiesRemoved,
+        AuthenticatorsAdded: AuthenticatorsAdded,
+        AuthenticatorsRemoved: AuthenticatorsRemoved,
+        MetadataURIUpdated: MetadataURIUpdated,
+        DAOURIUpdated: DAOURIUpdated,
```
pscott: camelcase should be `DaoUriUpdated` and `MetadataUriUpdated`

starknet/src/space/space.cairo
```diff
@@ -118,88 +119,135 @@ mod Space {
...
+    struct SpaceCreated {
+        space: ContractAddress,
+        owner: ContractAddress,
+        min_voting_duration: u32,
+        max_voting_duration: u32,
+        voting_delay: u32,
+        proposal_validation_strategy: Strategy,
+        proposal_validation_strategy_metadata_URI: Span<felt252>,
+        voting_strategies: Span<Strategy>,
+        voting_strategy_metadata_URIs: Span<Array<felt252>>,
```
pscott, line 153: should this be a span of spans?

starknet/src/space/space.cairo
pscott: yeah kinda sucks. I wonder why the `write` to storage function takes ownership of the data. Would be better if it were using a `@`...

pscott (approved):
ok syntax updated, now we need to submit a new PR to actually try these events :p

starknet/src/space/space.cairo
pscott: don't think it's needed to transform this into a span of spans. nvm this comment.

PR #526, opened 2023-08-30
add IQuorum

### LINKED ISSUE #523: expose quorum

pscott:
Not sure if the `IQuorum` interface should be in the folder `interfaces` or if it should be elsewhere. Thoughts @Orland0x ?

Closes #523

### DIFF
--- a/starknet/src/execution_strategies/simple_quorum.cairo
+++ b/starknet/src/execution_strategies/simple_quorum.cairo
@@ -20,6 +20,11 @@ mod SimpleQuorumExecutionStrategy {
     }
 
     #[internal]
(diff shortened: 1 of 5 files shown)


Orland0x:
Should we expose both quorum and proposal status in a single interface for `SimpleQuorum`?

pscott:
Yeah I don't know if it should be in the same interface. I think we should merge this one and discuss about it in https://github.com/snapshot-labs/sx-starknet/pull/527

Orland0x:
> Yeah I don't know if it should be in the same interface. I think we should merge this one and discuss about it in #527

okay

Orland0x (approved):
uTACK

PR #528, opened 2023-08-30
use correct syntax for no_update

### LINKED ISSUE #471: Use NoUpdateTrait instead of NoUpdate*
Once https://github.com/starkware-libs/cairo/issues/3678 is fixed, we can simply import NoUpdateTrait and just go for:
```rust
if input.voting_delay.should_update() {
 // code
}
```

instead of 
```rust
if NoUpdateU64::should_update(@input.voting_delay) {
  // code
}
```

pscott:
Finally!

Closes #471


Orland0x (approved):
uTACK

PR #534, opened 2023-09-06
feat: simple majority execution

### LINKED ISSUE #531: SimpleMajority execution strategy
Execution strategy that does not execute anything, but simply checks that `min_end_timestamp` has elapsed and that `For > Against`

pscott:
Closes #531


starknet/src/execution_strategies/no_execution_simple_majority.cairo
```diff
@@ -0,0 +1,47 @@
+/// Execution strategy that will not execute anything but ensure that the
+/// proposal is in the status `Accepted` or `VotingPeriodAccepted` by following
+/// the `SimpleMajority` rule (`votes_for > votes_against`).
+#[starknet::contract]
+mod NoExecutionSimpleMajorityExecutionStrategy {
+    use sx::interfaces::{IExecutionStrategy, IQuorum};
+    use sx::types::{Proposal, ProposalStatus};
+    use sx::execution_strategies::simple_quorum::SimpleQuorumExecutionStrategy;
```
Orland0x: not needed

starknet/src/utils/simple_majority.cairo
```diff
@@ -0,0 +1,179 @@
...
+    let timestamp = info::get_block_timestamp().try_into().unwrap();
+    if *proposal.finalization_status == FinalizationStatus::Cancelled(()) {
+        ProposalStatus::Cancelled(())
+    } else if *proposal.finalization_status == FinalizationStatus::Executed(()) {
+        ProposalStatus::Executed(())
+    } else if timestamp < *proposal.start_timestamp {
+        ProposalStatus::VotingDelay(())
+    } else if timestamp < *proposal.min_end_timestamp {
+        ProposalStatus::VotingPeriod(())
+    } else if timestamp < *proposal.max_end_timestamp {
```
Orland0x, line 19: we shouldnt have a `VotingPeriodAccepted` status with the strategy

Orland0x: i think just ignore `min_end_timestamp`

starknet/src/execution_strategies/no_execution_simple_majority.cairo
```diff
@@ -0,0 +1,47 @@
+/// Execution strategy that will not execute anything but ensure that the
+/// proposal is in the status `Accepted` or `VotingPeriodAccepted` by following
+/// the `SimpleMajority` rule (`votes_for > votes_against`).
+#[starknet::contract]
+mod NoExecutionSimpleMajorityExecutionStrategy {
+    use sx::interfaces::{IExecutionStrategy, IQuorum};
```
Orland0x: `IQuorum` not needed

starknet/src/utils/simple_majority.cairo
pscott: Hmm so the rationale behind that was:
- those who don't want the early end can set min/max to the same value
- those who want to have an early end (for whatever reason) can still have it without having to write another strategy

Orland0x: the thing is min/max are space params they cant be set just for proposals with this strategy. 

The goal of a 'simple majority' strategy is to find the majority of VP over the voting period. i dont think it would ever make sense to have an early end. So if spaces are using this execution strategy in combination with others, they would need to update space params often which would be annoying/ error prone.

pscott: fair enough ! f6ac7e04a8b18dc21688712bcd7acb7a1a1a2dcb

Orland0x (approved):
uTACK looks good

PR #543, opened 2023-09-08
refactor: internal impls

Orland0x:
moved internal functions to `InternalImpl` blocks, and removed their underscore prefixes. 

Removed Documentation style comments on internal functions


starknet/src/authenticators/stark_sig.cairo
```diff
@@ -1,7 +1,6 @@
 use starknet::ContractAddress;
 use sx::types::{Strategy, IndexedStrategy, Choice};
 
-/// See https://community.starknet.io/t/snip-off-chain-signatures-a-la-eip712/98029
```
pscott, line 4: would've been nice to keep this link somewhere in the codebase maybe for auditors?

pscott (approved):
doesn't reduce LOC but at least should help with "cleanliness" :D

starknet/src/authenticators/stark_sig.cairo
Orland0x: its in the stark eip712 lib still

PR #547, opened 2023-09-09
chore: move hardhat tests to root 

Orland0x:
didnt make sense for them to be in the /starknet sub dir


hardhat.config.ts
```diff
@@ -63,7 +63,7 @@ const config: HardhatUserConfig = {
   paths: {
     starknetSources: './starknet',
     sources: 'ethereum/src/',
-    tests: './starknet/test',
+    tests: './starknet/tests',
```
pscott: I just don't understand why these are in `starknet/tests` instead of `tests` ?

pscott (approved):

hardhat.config.ts
Orland0x: ah yeah should be `tests` doesnt look like thats needed though as they ran fine

Orland0x: updated

PR #548, opened 2023-09-12
refactor(eth sig): move chain id to message payload

Orland0x:


starknet/src/utils/eip712.cairo
```diff
@@ -105,6 +101,7 @@ mod EIP712 {
         ) -> u256 {
             let encoded_data = array![
                 u256 { low: PROPOSE_TYPEHASH_LOW, high: PROPOSE_TYPEHASH_HIGH },
+                Felt252IntoU256::into(starknet::get_tx_info().unbox().chain_id),
```
pscott, line 104:
```suggestion
                starknet::get_tx_info().unbox().chain_id.into(),
```

would that work?

tests/eth-sig-auth.test.ts
```diff
@@ -90,6 +89,7 @@ describe('Ethereum Signature Authenticator', function () {
 
     // PROPOSE
     const proposeMsg: Propose = {
+      chainId: '0x534e5f474f45524c49',
```
pscott, line 92: might be worth defining a constant `SN_GOERLI = ...` ?

pscott (approved):
utACK

starknet/src/utils/eip712.cairo
Orland0x: No because theres a namespace collision between the generic impl and the core lib one

pscott: love it :sad-pepe:

PR #549, opened 2023-09-13
refactor: array hashing

Orland0x:
Appends length in the `LegacyHashSpanFelt252` and removes it from `StructHashSpanFelt252`, which then becomes a simple wrapper around `LegacyHashSpanFelt252`.


pscott (approved):
utACK

PR #551, opened 2023-09-21
refactor: timestamp remapping

Orland0x:
remapped timestamps are now cached so only one call to the remapper needs to be made. This call should occur after the voting period has begun but before the first vote is cast.


starknet/src/utils/single_slot_proof.cairo
```diff
@@ -11,7 +11,51 @@ mod SingleSlotProof {
...
+        /// Queries the Timestamp Remapper contract for the closest L1 block number that occured before
+        /// the given timestamp and then caches the result. If the queried timestamp is less than the earliest
+        /// timestamp or larger than the latest timestamp in the mapper then the transaction will revert.
+        /// This function should be used to cache a remapped timestamp before its used when calling the 
+        /// `get_storage_slot` function with the same timestamp.
+        ///
+        /// # Arguments
+        ///
+        /// * `timestamp` - The timestamp at which to query.
+        /// * `tree` - The tree proof required query the remapper.
```
pscott:
```suggestion
        /// * `tree` - The tree proof required to query the remapper.
```

starknet/src/utils/single_slot_proof.cairo
```diff
@@ -11,7 +11,51 @@ mod SingleSlotProof {
     #[storage]
     struct Storage {
         _timestamp_remappers: ContractAddress,
```
pscott, line 13: do you know why it's `remapperS` if it's a single contract? is it a contract consisting of multiple contracts?

pscott (approved):
utACK

Orland0x:
will actually wait for some more updates from herodotus before merging this (if possible for audit fixes deadline)

starknet/src/utils/single_slot_proof.cairo
Orland0x: there can technical me multiple mappings in the contract i think, not sure when that would be useful though.

PR #573, opened 2023-10-04
N-08: fix misleading comments

### LINKED ISSUE #570: audit: N-08 Misleading Comments
The following comments are misleading or imprecise:
- This comment says "is not pending" when it should say "is pending".
- This error message does not match the condition it is checking, since the "caller" is the
authenticator contract.
Consider updating these comments accordingly.

pscott:
Closes #570


Orland0x (approved):
uTACK

Issue #554, opened 2023-10-04
audit: H-01 Inverted Case Style


Orland0x:
The StarkEIP712 contract supports accounts that use both snake and camel case.
However, the branch is inverted, so it attempts to use camel case on snake-case accounts and
vice versa. Importantly, it checks the ERC165 interface with the opposite case, which means
the verify_signature function will always fail, effectively disabling the
StarkSigAuthenticator .

PR #574, opened 2023-10-04
fix: stark sig auth account interaction

Orland0x:
To simply user actions, we removed support for alternate account implementations and now only support the standard introduced by SNIP-6.

this removes issue #554


starknet/src/tests/mocks/account.cairo
```diff
@@ -0,0 +1,253 @@
+// SPDX-License-Identifier: MIT
```
Orland0x: im not sure whether there is a way to add a contract in a dependency to the compilation artifacts. So just directly added the account  code here.

Orland0x: figured out how to do it

starknet/src/interfaces/i_account.cairo
```diff
@@ -15,18 +15,3 @@ trait AccountABI<TState> {
     fn is_valid_signature(self: @TState, hash: felt252, signature: Array<felt252>) -> felt252;
```
pscott: could we use the OZ library for the account interface? Or maybe there's a SNIP library somewhere we can use?

pscott (approved):
utACK

feelsgoodman.jpg

starknet/src/interfaces/i_account.cairo
Orland0x: yep imported from oz

PR #575, opened 2023-10-04
fix: C-02 Merkle Leaf not Authenticated

### LINKED ISSUE #553: audit: C-02 Merkle Leaf Not Authenticated
The MerkleWhitelistVotingStrategy contract is configured with a Merkle root that
represents a set of users and their voting power. When voting or proposing, users should
provide their leaf node along with a valid Merkle proof. However, the get_voting_power
function simply confirms that the leaf exists in the tree but does not validate that it belongs to
the voter. Therefore, anyone can use the voting power of any voter.
Consider validating that the leaf matches the expected voter.

pscott:
Authenticates the user when using the merkle whitelist authenticator.
Also adds a test case.

Closes #553


Orland0x (approved):
uTACK

PR #576, opened 2023-10-04
fix: L-01 Unused Strategy Slot

### LINKED ISSUE #558: audit: L-01 Unused Strategy Slot
The number of voting strategies per space contract should be limited to 256, so a single 256
bitmap can represent them all. This means the last available index should be 255. However, it
is actually limited to strictly less than 255. As an example, if the next index is 255 and there is
one voting strategy to add, this should be valid, but the assertion will fail since both sides of
the inequality are 255. Consider updating the range check to use an inclusive bound (i.e., <= ).

pscott:
Closes #558


Orland0x (approved):
uTACK

PR #577, opened 2023-10-04
fix: C-01 Votes Can Be Overcounted

### LINKED ISSUE #552: audit: C-01 Votes Can Be Overcounted
When making a proposal or voting on a proposal, the total voting power across multiple
strategies is combined. In both cases, the strategies are deduplicated.
This is achieved by incrementally constructing a bitmap with all the strategy indices. Each
index should be checked against the bitmap before being included. However, the deduplication
condition only checks index 0, so it is possible for any other strategy index to be repeated. This
means that the voting power for all other strategies can be counted multiple times.
Consider ensuring the deduplication condition covers all indices. In particular, it should assert
that (bit_map & s) == 0_u256 .

Orland0x:
closes #552

### DIFF
--- a/starknet/src/types/indexed_strategy.cairo
+++ b/starknet/src/types/indexed_strategy.cairo
@@ -41,6 +41,8 @@ impl IndexedStrategyImpl of IndexedStrategyTrait {
 
 #[cfg(test)]
 mod tests {
+    use core::array::ArrayTrait;
+    use core::traits::TryInto;
     use super::{IndexedStrategy, IndexedStrategyTrait};
 
     #[test]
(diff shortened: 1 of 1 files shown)


starknet/src/types/indexed_strategy.cairo
```diff
@@ -38,3 +38,39 @@ impl IndexedStrategyImpl of IndexedStrategyTrait {
...
+    #[test]
+    #[available_gas(100000000)]
+    fn empty_array() {
+        array![].span().assert_no_duplicate_indices();
+    }
+
+    #[test]
+    #[available_gas(100000000)]
+    #[should_panic(expected: ('Duplicate Found',))]
+    fn catch_duplicates() {
```
pscott, line 67: for the sake of completeness we might want to add a test where the duplicates are the first and last (256th) elements

pscott (approved):
utACK

starknet/src/types/indexed_strategy.cairo
Orland0x: added https://github.com/snapshot-labs/sx-starknet/pull/577/commits/2f674d84a7e87907bb58d156fbe16e30310584db

PR #579, opened 2023-10-05
fix: L-05 Infinite loop

### LINKED ISSUE #562: audit: L-05 Infinite Loop
The into_le_u64_array breaks the loop after processing the last element but does not
break when the loop is empty. This means that if the input array is empty, the function enters
an infinite loop.
Consider raising an error in this case.

pscott:
Panic if the array is empty

Closes #562

### DIFF
--- a/starknet/src/utils/endian.cairo
+++ b/starknet/src/utils/endian.cairo
@@ -38,7 +38,7 @@ fn into_le_u64_array(self: Array<u256>) -> (Array<u64>, u64) {
                 out.append(low_low);
             },
             Option::None => {
-                break (0);
+                panic_with_felt252('Array should not be empty!');
             }
         };
     };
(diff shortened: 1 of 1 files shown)


starknet/src/utils/endian.cairo
```diff
@@ -37,7 +37,9 @@ fn into_le_u64_array(self: Array<u256>) -> (Array<u64>, u64) {
                 out.append(low_high);
                 out.append(low_low);
             },
-            Option::None => {}
+            Option::None => {
+                break (0);
```
Orland0x: shouldn't we panic here?

pscott: i don't think so: if it's `None` it means `self` is empty.
we will return `out` which will be empty also, so we've turned an empty array into another empty array, which sounds fine to me?

Orland0x: okay yeah its probably okay then. 

Although this function has a specific usecase and should never be passed with an empty array.  The auditors also recommended to throw an error so might as well just to be safe.

pscott: Ok sir, your wish is my command

fd602371d5c0aafde565efd771d1b59685294599

Orland0x (approved):
uTACK

PR #580, opened 2023-10-05
fix: H-03 Incorrect Slot Key

### LINKED ISSUE #556: audit: H-03 Incorrect Slot Key
The get_mapping_slot_key function accepts a key and slot index in that order, but the
get_storage_slot function passes the parameters in reverse order. This means it will
calculate the wrong slot key and the storage proof will fail. Consequently, the
EthBalanceOfVotingStrategy cannot retrieve the voter's balance, making the strategy
unusable.
Consider passing the supplied arguments in the same order as the function signature.

pscott:
Use the correct argument order

Fixes #556


Orland0x (approved):
uTACK

PR #581, opened 2023-10-05
fix: M-01 Space Deployment vulnerable to front-running

### LINKED ISSUE #557: audit: M-01 Space Deployment Vulnerable To Front- Running
The deploy function of the factory is meant to deploy new spaces. However, it can be front-
run with the same class hash and salt but different initialization calldata. As a result, the attacker may set themselves as the owner of the victim's newly deployed space. The victim may miss or ignore the error messages, think that the space was deployed correctly and then proceed to use it. Consider adding the caller address to the salt in order to bind the contract address to the
deployer.

pscott:
Add the caller address to the salt computation

Closes #557

### DIFF
--- a/starknet/src/factory/factory.cairo
+++ b/starknet/src/factory/factory.cairo
@@ -40,8 +40,14 @@ mod Factory {
             contract_address_salt: felt252,
             initialize_calldata: Span<felt252>
         ) -> SyscallResult<ContractAddress> {
+            // We create the salt by hashing the user provided salt and the caller address
+            // to avoid any frontrun attacks.
+            let caller_address: felt252 = starknet::info::get_caller_address().into();
+            let salt_input = array![caller_address, contract_address_salt];
+            let salt = poseidon::poseidon_hash_span(salt_input.span());
+
             let (space_address, _) = syscalls::deploy_syscall(
-                class_hash, contract_address_salt, array![].span(), false
+                class_hash, salt, array![].span(), false
             )?;
 
             // Call initializer. 
--- a/starknet/src/factory/factory.cairo
+++ b/starknet/src/factory/factory.cairo
@@ -6,7 +6,7 @@ trait IFactory<TContractState> {
         ref self: TContractState,
         class_hash: ClassHash,
         initialize_calldata: Span<felt252>,
-        contract_address_salt: felt252,
+        salt_nonce: felt252,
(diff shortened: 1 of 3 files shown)


starknet/src/factory/factory.cairo
```diff
@@ -40,8 +40,14 @@ mod Factory {
             contract_address_salt: felt252,
```
Orland0x: we should probably call this a `salt_nonce` now. 

also might as well make it more equivalent to sx-evm by placing the nonce at the end of the args:  https://github.com/snapshot-labs/sx-evm/blob/df37a70673067300a3a6c0541740117a86d0bc59/src/ProxyFactory.sol#L12C86-L12C86

starknet/src/factory/factory.cairo
```diff
@@ -40,8 +40,14 @@ mod Factory {
...
             initialize_calldata: Span<felt252>
         ) -> SyscallResult<ContractAddress> {
+            // We create the salt by hashing the user provided salt and the caller address
+            // to avoid any frontrun attacks.
+            let caller_address: felt252 = starknet::info::get_caller_address().into();
```
Orland0x: no type inference?

pscott: b824574e6ca0e542fe5d61ac1963005a6d7e662b

starknet/src/factory/factory.cairo
pscott: yeah original naming was based on the [deploy_syscall argument name](https://github.com/starkware-libs/cairo/blob/775b4f84e705293ded7b7cc203650eb983246842/corelib/src/starknet/syscalls.cairo#L24).. `salt_nonce` sounds weird to me because `nonce` refers to something we increment usually no (I see that's what we have on the EVM side of things but wonder where the name comes from)


how about just `salt`? :p

Orland0x: yeah thats probably what we would do right? if a user wanted to deploy an identical space they would just need to increment the nonce

pscott: yeah I guess so. In my mind we would just query a random number, because incrementing a nonce requires knowing which nonce was previously used, which sounds more cumbersome than just genereating a random number

ab1004cee63ddf6ffb008a1a36a7109e451f1e62

Orland0x (approved):
uTACK

PR #582, opened 2023-10-05
fix: N-07 Duplicate Code

### LINKED ISSUE #569: audit: N-07 Duplicate Code
The default contract address is implemented in user_address.cairo and
proposal.cairo .
Consider removing the redundant version.

pscott:
Closes #569

### DIFF
--- a/starknet/src/types/default.cairo
+++ b/starknet/src/types/default.cairo
@@ -1,0 +1,8 @@
+use starknet::{ContractAddress, contract_address_const};
+
+// Required for the proposal derivation. Ideally, ContractAddress would impl Default in the corelib.
+impl ContractAddressDefault of Default<ContractAddress> {
+    fn default() -> ContractAddress {
+        contract_address_const::<0>()
+    }
+}
--- a/starknet/src/types/proposal.cairo
+++ b/starknet/src/types/proposal.cairo
@@ -1,7 +1,7 @@
 use starknet::{ContractAddress, storage_access::StorePacking, Store, contract_address_const};
 use sx::{
-    utils::math::pow,
-    types::{ContractAddressDefault, FinalizationStatus, UserAddress, UserAddressTrait},
+    utils::math::pow, types::{FinalizationStatus, UserAddress, UserAddressTrait},
+    utils::ContractAddressDefault,
 };
 
 const BITMASK_32: u128 = 0xffffffff;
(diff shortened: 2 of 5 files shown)


starknet/src/types/proposal.cairo
```diff
@@ -1,6 +1,9 @@
 use starknet::{ContractAddress, storage_access::StorePacking, Store, contract_address_const};
 use sx::{
-    utils::math::pow, types::{FinalizationStatus, UserAddress, user_address::UserAddressTrait}
+    utils::math::pow,
+    types::{
+        FinalizationStatus, UserAddress, user_address::{UserAddressTrait, ContractAddressDefault}
```
Orland0x: we re-exported `user_address` so shouldnt need the full import path

starknet/src/types/proposal.cairo
```diff
@@ -43,13 +46,6 @@ struct Proposal {
     active_voting_strategies: u256
 }
 
-// Required for the proposal derivation. Ideally, ContractAddress would impl Default in the corelib.
-impl ContractAddressDefault of Default<ContractAddress> {
```
Orland0x, line 47: might make more sense to put this in a dedicated file like `defaults.cairo` in `/utils` as its kind of random here. We have `into.cairo` which  serves similar role for the `Into` trait (impls of standard lib types that are not (yet) in the standard lib).

starknet/src/types/proposal.cairo
pscott: 022db4eec4c3cb91d24b7f99eb679ea1771c2039

starknet/src/types/proposal.cairo
pscott: true! as I said, ideally, ContractAddress would have a default impl in core lol

7acb81eaa22fab7371b982c8da3358fc70c570d4

Orland0x: agreed. `defaults.cairo` is in `types` not `utils` also dont think the impl has been removed from `user_address.cairo`

starknet/src/types/default.cairo
```diff
@@ -0,0 +1,8 @@
+use starknet::{ContractAddress, contract_address_const};
+
+// Required for the proposal derivation. Ideally, ContractAddress would impl Default in the corelib.
```
Orland0x: comment no longer really applicable

starknet/src/types/proposal.cairo
pscott: 1872fed814ed4136e3d6e63eb282a2a3ae0a3ea6

starknet/src/types/default.cairo
pscott: 1872fed814ed4136e3d6e63eb282a2a3ae0a3ea6

Orland0x (approved):
uTACK

PR #583, opened 2023-10-05
fix: N-01 Inconsistent storage pattern

### LINKED ISSUE #563: audit: N-01 Inconsistent Storage Pattern
The codebase mostly does not declare storage variables that are not specifically required, even
if they are implicitly part of the storage. For example, the Space storage struct does not
include an owner, even though it is implicitly referenced. However, the
EthSigAuthenticator contract does declare a _domain_hash variable that is only
implicitly referenced within the contract.
Consider choosing one of the patterns and applying it consistently throughout the codebase.

pscott:
Remove the unused `_domain_hash` storage vars

Closes #563

### DIFF
--- a/starknet/src/authenticators/eth_sig.cairo
+++ b/starknet/src/authenticators/eth_sig.cairo
@@ -93,7 +93,6 @@ mod EthSigAuthenticator {
 
     #[storage]
     struct Storage {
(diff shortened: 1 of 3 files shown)


Orland0x (approved):
uTACK

PR #584, opened 2023-10-06
fix: L-04 Risky upgrade pattern

### LINKED ISSUE #561: audit: L-04 Risky Upgrade Pattern
The space is upgraded by replacing the class hash and then calling the new initializer.
However, the new initializer has to be designed specifically as an upgrade function, with
reference to the current contract state. In particular, if the new initializer is similar to the current
one, it would reset the next proposal id. It would also require the upgrade to include new voting
strategies and authenticators. On the other hand, if it is designed as an upgrader and does not
affect these parameters, it would not function as a complete initializer if the new contract was
deployed independently.
Instead of calling the new initializer, consider calling a separate single-use upgrade function
that accounts for the current contract state. Alternatively, consider using the versioned
initializer pattern.

pscott:
Adds a new function `post_upgrade_initializer` that is meant to be called by the `upgrade` function (previously, it called the `initialize` function).

Name can be changed, if something better comes to mind?

Closes #561


Orland0x (approved):
uTACK

starknet/src/interfaces/i_space.cairo
```diff
@@ -151,4 +151,13 @@ trait ISpace<TContractState> {
...
     ) -> SyscallResult<()>;
+
+    /// Initializes the contract after an upgrade. This is different from the `initialize` function
+    /// in that it is called after the upgrade, which means the contract already has a state and storage
+    /// associated to it.
+    ///
+    /// # Arguments
+    ///
+    /// * `initialize_calldata` - The calldata to use to perform a post-upgrade initialization.
+    fn post_upgrade_initializer(ref self: TContractState, initialize_calldata: Array<felt252>,);
```
Orland0x, line 162: do we need this function in the interface for the V1?

pscott: felt clean to have it even if it doesn't do anything.. makes it clearer for v2 imho but no strong opinion

PR #587, opened 2023-10-06
fix: N-02 Code Clarity Suggestions

### LINKED ISSUE #564: audit: N-02 Code Clarity Suggestions
- When reading felt252 arrays from storage, the pointer offset is increased by the
felt252 size, but the exit condition does not account for size. This behaves correctly
because the size is actually 1, but we believe it would still be clearer to include it
consistently.
- This hardcoded constant could use BoundedU8::max for readability.
- This hardcoded address could use zero for readability.
- This addition is unnecessary because it is equivalent to with_prefix = u256 {low:
input, high: prefix} .
- The MerkleWhitelistVotingStrategy contract seems to unnecessarily split the
user_params so they can be deserialized independently, instead of deserializing them
together.
- The uint256_into_le_u64s function could return the components in a systematic
order (i.e., most-significant 64-bit value to least-significant 64-bit value).
- This error message could be changed to Already finalized for consistency with
the rest of the codebase.
- These two checks ([1] and [2]) can leverage the is_non_zero function.

pscott:
- Use `felt252` size and `u8` size in the read/write storage code
- Use `BoundedU8` instead of `255_u8`
- Use `ContractAddressZeroable` for `ContractAddressDefault`
- Remove useless addition of `with_prefix`
- Simplify the deserialization in `MerkleWhitelistVotingStrategy`
- Return the components in a systematic order in `uint256_into_le_u64s`
- [ ] waiting for more information regarding `Already Finalized` error message
- Use `is_non_zero()` when possible

Closes #564

### DIFF
--- a/starknet/src/authenticators/eth_tx.cairo
+++ b/starknet/src/authenticators/eth_tx.cairo
@@ -185,7 +185,7 @@ mod EthTxAuthenticator {
     impl InternalImpl of InternalTrait {
         fn consume_commit(ref self: ContractState, hash: felt252, sender_address: EthAddress) {
             let committer_address = self._commits.read(hash);
-            assert(!committer_address.is_zero(), 'Commit not found');
+            assert(committer_address.is_non_zero(), 'Commit not found');
             assert(committer_address == sender_address, 'Invalid sender address');
             // Delete the commit to prevent replay attacks.
             self._commits.write(hash, Zeroable::zero());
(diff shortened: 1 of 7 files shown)


Orland0x:
dude i just also did all of these ffs was about to push.

pscott:
lmao frontran

pscott:
thought u were working on https://github.com/snapshot-labs/sx-starknet/issues/555.. my bad, should've self assigned when i started. mb mb :(

Orland0x:
assume they meant the 'Proposal has been finalized' message as thats the only one not like the others. i fixed on my branch

pscott:
Superseeded by https://github.com/snapshot-labs/sx-starknet/pull/588

PR #588, opened 2023-10-06
refactor: N-02 Code Clarity Suggestions

### LINKED ISSUE #564: audit: N-02 Code Clarity Suggestions

Orland0x:
closes #564

### DIFF
--- a/starknet/src/types/strategy.cairo
+++ b/starknet/src/types/strategy.cairo
@@ -1,5 +1,4 @@
-use starknet::ContractAddress;
-use starknet::{StorageBaseAddress, Store, SyscallResult};
+use starknet::{ContractAddress, StorageBaseAddress, Store, SyscallResult};
 
 /// A strategy identified by an address
 #[derive(Clone, Drop, Option, Serde, starknet::Store)]
@@ -47,7 +46,7 @@ impl StoreFelt252Array of Store<Array<felt252>> {
         offset += 1;
 
         // Sequentially read all stored elements and append them to the array.
-        let exit = len + offset;
+        let exit = len * Store::<felt252>::size() + offset;
         loop {
             if offset >= exit {
                 break;
--- a/starknet/src/types/strategy.cairo
+++ b/starknet/src/types/strategy.cairo
@@ -1,3 +1,4 @@
+use integer::BoundedU8;
 use starknet::{ContractAddress, StorageBaseAddress, Store, SyscallResult};
 
 /// A strategy identified by an address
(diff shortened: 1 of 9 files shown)


starknet/src/voting_strategies/merkle_whitelist.cairo
```diff
@@ -32,15 +32,12 @@ mod MerkleWhitelistVotingStrategy {
             params: Span<felt252>, // [root: felt252]
             user_params: Span<felt252>, // [leaf: Leaf, proof: Array<felt252>]
         ) -> u256 {
-            let cache = user_params; // cache
+            let mut params = params;
```
pscott: you can add `mut` keywrods to the `params` and `user_params` arguments directly, no need to re-bind :D

starknet/src/types/strategy.cairo
```diff
@@ -47,7 +47,7 @@ impl StoreFelt252Array of Store<Array<felt252>> {
         offset += 1;
 
         // Sequentially read all stored elements and append them to the array.
-        let exit = len + offset;
+        let exit = len * Store::<felt252>::size() + offset;
```
pscott, line 50: I did this, what do you think https://github.com/snapshot-labs/sx-starknet/pull/587/commits/73a92691ab3739333bb96533d3aba68918516157 ?

pscott (approved):
utACK

starknet/src/voting_strategies/merkle_whitelist.cairo
Orland0x: oh nice, updated. think there some other situations where we can do this. will have a look

starknet/src/types/strategy.cairo
Orland0x: yeah thanks, that makes a lot of sense.

PR #589, opened 2023-10-06
refactor: N-06 Incomplete Encapsulation

### LINKED ISSUE #568: audit: N-06 Incomplete Encapsulation
For efficiency reasons, the StarknetSpaceManager uses uint256 values to internally
represent booleans. However, the uint256 is also returned from external queries, which
requires the caller to know the convention.
In the interest of predictability, consider returning a boolean when queried externally.

pscott:
Closes #568

### DIFF
--- a/ethereum/src/execution-strategies/StarknetSpaceManager.sol
+++ b/ethereum/src/execution-strategies/StarknetSpaceManager.sol
@@ -48,9 +48,9 @@ contract StarknetSpaceManager is OwnableUpgradeable {
 
     /// @notice Check if a space is enabled.
     /// @param space Address of the space.
-    /// @return uint256 whether the space is enabled.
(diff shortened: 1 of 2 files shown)


Orland0x:
do you think this is worth it considering it will then deviate from sx-evm? I assume the sx.js clients will use a lot of the same code for the solidity contracts in sx-sn and sx-evm particularly

pscott:
Brings the question: should we also update this on SX-EVM side maybe?

Orland0x:
> Brings the question: should we also update this on SX-EVM side maybe?

Dont think its really worth it post audit, feels like a minor/debatable thing. 

Its the responsibility of external integrators to check the interfaces of the contracts they interact with. its not like we are implementing some existing standard.

pscott:
Ok then happy to close this PR and add a comment for the auditors

PR #590, opened 2023-10-07
refactor: N-10 Naming Suggestions

### LINKED ISSUE #572: audit: N-10 Naming Suggestions
The following function variables could benefit from more descriptive naming:
- The proposal validation strategy validates that the author can make a proposal, rather
than validating anything about the proposal itself, and could be renamed accordingly.
- The update_proposal function could be renamed to
update_proposal_execution_strategy , since most of the proposal parameters
cannot be changed.
- The authenticator contracts could rename target to space in all functions to better
distinguish it from the proposal target.
- The l1_account_address variable does not refer to an account, and could be
renamed to l1_token_address or l1_contract_address .
- The reinitialize function merely prepares the contract for a future reinitialization,
and could be renamed to reset or allow_reinitialization .
- The Eth prefix in the EthBalanceOfVotingStrategy contract refers to Ethereum, but
could be mistaken for ETH (the asset). It could be renamed to
L1BalanceOfVotingStrategy .

pscott:
- [x] rename `validate` -> decided against, want to preserve SX-EVM compatibility and we feel like the wording is "ok" 
- [x] rename `update_proposal` -> decided against, as `update_proposal` can also update the metadata_uri, and I feel like it would be too verbose
- [x] rename `target` when appropriate https://github.com/snapshot-labs/sx-starknet/pull/590/commits/b9c2755d655820e2f86926400a54da7b2179908f
- [x] rename `l1_account_address` a571f1b728d8ac92369c11c7cb9996f0ccb1e1f5
- [x] rename `reinitialize` 5bb2b0100c0905ab4e72479c3a1e7c14fbf31a3f
- [x] rename `EthBalanceOfVotingStrategy` 4aa949e68cb02f2b38454e705285ce70db3ea80f && 74a7452490167be9a2f56b5de9ae0f151721da4c && https://github.com/snapshot-labs/sx-starknet/pull/590/commits/0b563e2c9f672b3afeb838bb1bac417dbac25fb4

Closes #572


starknet/src/voting_strategies/eth_balance_of.cairo
```diff
@@ -37,7 +37,7 @@ mod EthBalanceOfVotingStrategy {
 
             // Decode params 
             let mut params = params;
-            let (l1_account_address, slot_index) = Serde::<(
+            let (l1_token_address, slot_index) = Serde::<(
```
Orland0x: this can be used to prove a slot in any contract, so token is potentially not correct. The reason I used account as thats whats they are referred to in the EVM spec, and as this contract is dealing with evm internals it made sense to follow that as close as possible. I can see how this could be confused with starknet accounts though, so `l1_contract_address` probably makes the most sense.

pscott: Should we reconsider the wording of `balance` then maybe? Because balance seems to refer to a token (erc20 or nft). wdyt?

Orland0x: oh got confused thought we were looking at `single_slot_proof.cairo` library which does indeed use `l1_contract_address` already. https://github.com/snapshot-labs/sx-starknet/blob/4aa949e68cb02f2b38454e705285ce70db3ea80f/starknet/src/utils/single_slot_proof.cairo#L31

This wrapper voting strategy i intended to make more specific to tokens hence calling it `eth_balance_of` even though it can technically be used to prove any slot as its easier to understand for people not super familiar with storage proofs etc.

bonustrack: > Should we reconsider the wording of `balance` then maybe? Because balance seems to refer to a token (erc20 or nft). wdyt?

It may not be a "balance" per say, storage proof can be used to beyond token / NFT holding

Orland0x:
should we rename `eth_tx` to `l1_tx`? Might be worth being consistent to remove all mentions of 'eth' unless it specifically refers to the token itself

starknet/src/voting_strategies/eth_balance_of.cairo
```diff
@@ -1,5 +1,5 @@
 #[starknet::contract]
```
Orland0x, line 1: should change filename `eth_balance_of.cairo` too

starknet/src/voting_strategies/eth_balance_of.cairo
pscott: Yeah I'm torn between using something "token specific" so users reading code would not be lost, and calling it something general like "L1SlotValue" ...

pscott:
> should we rename `eth_tx` to `l1_tx`? Might be worth being consistent to remove all mentions of 'eth' unless it specifically refers to the token itself

I don't know about that, if I say out loud "you can authenticate via an eth signature" or "you can authenticate via an l1 signature" I feel like the former is clearer ?

starknet/src/voting_strategies/eth_balance_of.cairo
bonustrack: We shouldn't use token IMO, naming should be specific to what the code is doing. For users we can always name things differently in the UI, slot value is the one making more sense to me (would remove L1 as it may not even be L1 but others L2 in future).

Orland0x:
> > should we rename `eth_tx` to `l1_tx`? Might be worth being consistent to remove all mentions of 'eth' unless it specifically refers to the token itself
> 
> I don't know about that, if I say out loud "you can authenticate via an eth signature" or "you can authenticate via an l1 signature" I feel like the former is clearer ?

yeah true

starknet/src/voting_strategies/eth_balance_of.cairo
pscott: 74a7452490167be9a2f56b5de9ae0f151721da4c

starknet/src/voting_strategies/eth_balance_of.cairo
pscott: Decided to rename it to `EvmSlotValueVotingStrategy`, as it is the one that makes most sense. I added, as a note, that this is often used for computing a L1 token balance, or NFT balance :) 0b563e2c9f672b3afeb838bb1bac417dbac25fb4

Orland0x (approved):
uTACK

PR #593, opened 2023-10-09
fix: H-02 Votes Can Be Blocked

### LINKED ISSUE #555: audit: H-02 Votes Can Be Blocked
The StarknetCommit contract allows anyone to submit a hash representing their vote,
which will be recorded with their L1 address in the EthTxAuthenticator contract. Later,
they can consume the commitment, provided the hash matches a valid transaction (i.e.,
proposal, vote, or proposal update) from the original sender.
However, an attacker can submit a hash corresponding to another user's valid transaction. This
will prevent the victim from making the same commitment to execute the transaction.
Fortunately, since metadata_uri is a free parameter, they could commit to an equivalent
transaction with a different metadata_uri value, and use a private relayer to ensure they
cannot be front-run.
Consider including the L1 message sender as part of the hash commitment to ensure
uniqueness. Alternatively, consider using a two-dimensional storage mapping to allow multiple
users to commit to the same hash.

Orland0x:
Replaces the `_commits` (commit -> address) mapping with a double mapping (commit, address -> bool). 

This allows the same commit to be made by 2 different addresses, which wasn't possible before. 

This closes #555 as now its impossible for a user to block another by making a valid commit for their action before they do.


tests/eth-tx-auth.test.ts
```diff
@@ -473,7 +473,7 @@ describe('Ethereum Transaction Authenticator', function () {
       );
       expect.fail('Should have failed');
     } catch (err: any) {
-      expect(err.message).to.contain(shortString.encodeShortString('Invalid sender address'));
+      expect(err.message).to.contain(shortString.encodeShortString('Commit not found'));
     }
```
pscott, line 477: could we add a test that tests the "attack" vector that OZ mentioned? Just to add it as a regression test for later work

pscott (approved):
utACK, any reason you chose double mapping instead of adding the l1 message sender address?

tests/eth-tx-auth.test.ts
Orland0x: yeah sorry added but forgot to push

Orland0x:
> utACK, any reason you chose double mapping instead of adding the l1 message sender address?

I had a think about the other approach but not sure its really feasible as it would mean computing the hash on-chain in order to include `msg.sender`. We use posiodon which isnt really possible to do on L1. Could switch to keccak but then it would be expensive to verify on Starknet

PR #598, opened 2023-10-20
Lock post_upgrade_initializer

pscott:
Adds a comment and some lock function to ensure post_upgrade_initializer is only called once

### DIFF
--- a/starknet/src/space/space.cairo
+++ b/starknet/src/space/space.cairo
@@ -481,7 +481,6 @@ mod Space {
             // you might not want to re-initialize the WHOLE contract state!
             let mut state = Reinitializable::unsafe_new_contract_state();
             Reinitializable::InternalImpl::initialize(ref state);
-
         // This contract being the first version, we don't expect anyone to upgrade to it.
         // We leave the implementation empty.
         }
--- a/starknet/src/space/space.cairo
+++ b/starknet/src/space/space.cairo
@@ -477,8 +477,11 @@ mod Space {
         fn post_upgrade_initializer(ref self: ContractState, initialize_calldata: Array<felt252>,) {
             // This code is left here to indicate to future developers that this
             // function should be called only once!
-            let state = Reinitializable::unsafe_new_contract_state();
-            Reinitializable::InternalImpl::not_initialized(@state);
+            // Note: the `initialize` function is probably NOT the function to call here, as
+            // you might not want to re-initialize the WHOLE contract state!
+            let mut state = Reinitializable::unsafe_new_contract_state();
+            Reinitializable::InternalImpl::initialize(ref state);
+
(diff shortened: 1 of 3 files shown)


starknet/src/space/space.cairo
```diff
@@ -474,9 +474,12 @@ mod Space {
...
         }
 
-        fn post_upgrade_initializer(
-            ref self: ContractState, initialize_calldata: Array<felt252>,
-        ) { // This contract being the first version, we don't expect anyone to upgrade to it.
+        fn post_upgrade_initializer(ref self: ContractState, initialize_calldata: Array<felt252>,) {
+            // This code is left here to indicate to future developers that this
+            // function should be called only once!
+            let state = Reinitializable::unsafe_new_contract_state();
+            Reinitializable::InternalImpl::initialized(@state);
```
Orland0x: this should call `initialize` instead right?

pscott: yeah i didn't know how to convey the idea exactly... if we call `initialize` then the user won't necessarily understand that it should check that it hasn't already been initialized. Also, the work `initialize` may be confused with the already-existing `initialize` function, which should not be called here (it's the whole idea of `post_upgrade_initialize`...)

Orland0x: `initialize` internally checks that it has not been initialized already: https://github.com/snapshot-labs/sx-starknet/blob/4f602aa269ef92e279a346fc029fdebcc516a322/starknet/src/utils/reinitializable.cairo#L15C10-L15C10

Currently we cannot call `post_upgrade_initialize` after `upgrade` as `initialized` will always revert

pscott: Yes but as I said, I think here what's important is to show the *intent* of what we're doing. I know that `initalize` checks that it has not already been initialized but if we write `initialize` here then people might think "oh yeah we need to initialize here" when, in fact, it's not exactly that that needs to be done

pscott: If you think calling `initialize` makes it clearer, then let's go for it :)

pscott: I realized that my point was calling `not_iniitalized` instead of `initialized` :)

Orland0x: i see what you mean about `initialize` being confusing, ideally we would have used different naming for the `Reinitializable` module. 

But yeah to me `initialize` is the most clear. If the `post_upgrade_initialize` function was empty (eg space v2 added a new function but no state changes needed) then it could be left as it is.

pscott: ecfc53bb63bebff4c7ec73fc5e8b07f25d9c88aa added a comment to try to make it clear

Orland0x (approved):
Looks good. uTACK

PR #600, opened 2023-11-02
feat: starknet timelock execution strategy

Orland0x:
Starknet Timelock Execution Strategy and tests.


starknet/src/execution_strategies/timelock.cairo
```diff
@@ -0,0 +1,341 @@
...
+    #[constructor]
+    fn constructor(
+        ref self: ContractState,
+        owner: ContractAddress,
+        veto_guardian: ContractAddress,
+        spaces: Span<ContractAddress>,
+        timelock_delay: u32,
+        quorum: u256
+    ) {
+        // Migration to components planned ; disregard the `unsafe` keyword,
```
pscott, line 124: btw: https://github.com/OpenZeppelin/cairo-contracts/releases/tag/v0.8.0 ! Getting close :D

starknet/src/execution_strategies/timelock.cairo
```diff
@@ -0,0 +1,341 @@
...
+        fn execute(
+            ref self: ContractState,
+            proposal_id: u256,
+            proposal: Proposal,
+            votes_for: u256,
+            votes_against: u256,
+            votes_abstain: u256,
+            payload: Array<felt252>
+        ) {
+            let state = SpaceManager::unsafe_new_contract_state();
```
pscott, line 152: missing comment about `unsafe` here?

starknet/src/execution_strategies/timelock.cairo
```diff
@@ -0,0 +1,341 @@
...
+            SpaceManager::InternalImpl::assert_only_spaces(@state);
+            let state = SimpleQuorum::unsafe_new_contract_state();
+            let proposal_status = SimpleQuorum::InternalImpl::get_proposal_status(
```
pscott, line 155: should we call `self.get_proposal_status(...)` to limit code duplication ?

starknet/src/execution_strategies/timelock.cairo
pscott: and other places in the code, so I'm wondering if it's "on purpose" or not?

starknet/src/tests/execution_strategies/timelock.cairo
```diff
@@ -0,0 +1,591 @@
...
+                    calldata: array![],
+                    salt: 0
+                }
+            ]
+        );
+        // Proposal is not queued
+        testing::set_block_timestamp(
+            info::get_block_timestamp() + timelock.timelock_delay().into()
+        );
+        // Execute proposal on timelock - should fail because the timelock delay hasn't passed
```
pscott: comment is incorrect; should fail because proposal is not queued, not because delay hasn't passed

starknet/src/tests/execution_strategies/timelock.cairo
```diff
@@ -0,0 +1,591 @@
...
+        space.execute(proposal_id, payload.clone());
+        // Veto proposal from unauthorized account
+        testing::set_caller_address(timelock.owner());
+        testing::set_contract_address(timelock.owner());
+        timelock.veto(poseidon::poseidon_hash_span(payload.span()));
+        testing::set_block_timestamp(
+            info::get_block_timestamp() + timelock.timelock_delay().into()
+        );
+        // Execute proposal on timelock - should fail as proposal was vetoed
+        timelock.execute_queued_proposal(payload.span());
```
pscott: can remove this

starknet/src/utils/simple_quorum.cairo
```diff
@@ -33,10 +34,9 @@ mod SimpleQuorumExecutionStrategy {
             votes_against: u256,
             votes_abstain: u256,
         ) -> ProposalStatus {
-            let accepted = InternalImpl::quorum_reached(
-                self._quorum.read(), votes_for, votes_against, votes_abstain
-            )
-                & InternalImpl::supported(votes_for, votes_against);
+            let quorum_reached = votes_for + votes_abstain >= self._quorum.read();
```
pscott, line 37: we do no count the `votes_against` anymore? I believe this differs from the previous implem?

pscott: Actually, I'm surprised no test broke ^^ 
we should have a test with only `AGAINST` so that it reaches quorum but with no `abstain` or `for`

starknet/src/utils/space_manager.cairo
```diff
@@ -0,0 +1,62 @@
...
+        space: ContractAddress
+    }
+
+    #[generate_trait]
+    impl InternalImpl of InternalTrait {
+        fn initializer(ref self: ContractState, mut spaces: Span<ContractAddress>) {
+            loop {
+                match spaces.pop_front() {
+                    Option::Some(space) => {
+                        self._spaces.write(*space, true);
```
pscott, line 36: we do not check if `space.is_zero()` here, but we do in `enable_space`. thoughts?

pscott (changes requested):
Good work! A couple of comments :)

starknet/src/tests/execution_strategies/timelock.cairo
Orland0x: agreed

starknet/src/utils/space_manager.cairo
Orland0x: I kept it consistent with [sx-evm](https://github.com/snapshot-labs/sx-evm/blob/a17a4abbc8a35276163fbbb051bf14c879980b96/src/utils/SpaceManager.sol#L26) but that doesnt really make much sense. Think we should add the zero check.

starknet/src/utils/simple_quorum.cairo
Orland0x: Nope we dont add the against votes, same as [sx-evm](https://github.com/snapshot-labs/sx-evm/blob/a17a4abbc8a35276163fbbb051bf14c879980b96/src/execution-strategies/SimpleQuorumExecutionStrategy.sol#L72). 

Otherwise it would create an adverse incentive for those voting against to just not vote at all.

starknet/src/execution_strategies/timelock.cairo
Orland0x: No, that would call the external `get_proposal_status` function wheras we want to call the internal one within the `SimpleQuorum` module.

starknet/src/execution_strategies/timelock.cairo
Orland0x: yep not been fully consistent, will add missing ones

starknet/src/utils/space_manager.cairo
Orland0x: added the zero (and also duplicate) check, along with regression tests

pscott (approved):
utACK

PR #603, opened 2023-12-04
chore: herodotus integration testing

Orland0x:
Adds a demo script to test the proposal creation and voting flow with storage proofs. 

The single slot proof voting strategy required a few interface updates due to changes in the Herodotus interface. 

Additionally, there were external functions defined within `single_slot_proof.cairo` which were not actually exposed in the contract interface (discovered external functions defined in modules are not included in the contract interface). These were moved to `evm_slot_value.cairo` in order to be properly exposed.


pscott (approved):
well done :D

PR #604, opened 2023-12-08
chore: update confusing evm slot value slotIndex type annotation

Sekhmet:
It's decoded as `u256`, not `u32`.


Orland0x (approved):
nice spot. uTACK

PR #605, opened 2024-01-11
feat: support delegation with EvmSlotValueVotingStrategy

Orland0x:
Before, we were just using l1 token balances for voting power using the `EVMSlotValueVotingStrategy`. This did not allow delegation. 

This PR changes the voting strategy to now query the checkpoints array within the token state. The last checkpoint in this array for a particular voter contains their delegated voting power at the current block number. 

To obtain this voting power we can verify a storage proof of the storage slot where this checkpoint resides. 

However we must ensure that the slot submitted does indeed correspond to the final checkpoint in the voter's checkpoints array. Since array values are stored contiguously in the contract storage, we can do this by additionally verifying that the immediately following slot is empty.

Refering to the solidity [documentation](https://docs.soliditylang.org/en/v0.8.23/internals/layout_in_storage.html#mappings-and-dynamic-arrays) we can derive the slot key of the final checkpoint as: 

```
keccak256(abi.encode(keccak256(abi.encode(voterAddress, uint256(checkpointsMappingSlotIndex))))) + token.numCheckpoints(voterAddress) - 1
```

We now submit 2 storage proofs with each vote which are verified in succession. 

Here is an example vote on starknet goerli: [0x06efdb5b164cc77f03c8e692f457b1e27805ee7c59cdbfe7ce52eafd1775aa18](https://testnet.starkscan.co/tx/0x06efdb5b164cc77f03c8e692f457b1e27805ee7c59cdbfe7ce52eafd1775aa18)

### DIFF
--- a/scripts/deploy-space.ts
+++ b/scripts/deploy-space.ts
@@ -23,21 +23,25 @@ async function main() {
   const provider = new RpcProvider({ nodeUrl: starknetNetworkUrl });
   const account = new Account(provider, accountAddress, accountPk);
 
-  const l1TokenAddress = '0xd96844c9B21CB6cCf2c236257c7fc703E43BA071'; //OZ token 18 decimals
-  const slotIndex = cairo.uint256(8);
+  // OZ Votes token 18 decimals
+  const l1TokenAddress = '0xd96844c9B21CB6cCf2c236257c7fc703E43BA071'; 
 
+  // Slot index of the checkpoints mapping in the token contract, 
+  // obtained using Foundry's Cast Storage Layout tool. 
+  const slotIndex = cairo.uint256(8); 
+  
   const factsRegistryAddress = '0x01b2111317EB693c3EE46633edd45A4876db14A3a53ACDBf4E5166976d8e869d';
   const timestampsRemapperAddress =
     '0x2ee57d848297bc7dfc8675111b9aa3bd3085e4038e475250770afe303b772af';
 
   const evmSlotValueVotingStrategySierra = json.parse(
     fs
-      .readFileSync('starknet/target/dev/sx_EvmSlotValueVotingStrategy.sierra.json')
+      .readFileSync('starknet/target/dev/sx_OZVotesStorageProofVotingStrategy.sierra.json')
       .toString('ascii'),
   );
   const evmSlotValueVotingStrategyCasm = json.parse(
     fs
-      .readFileSync('starknet/target/dev/sx_EvmSlotValueVotingStrategy.casm.json')
+      .readFileSync('starknet/target/dev/sx_OZVotesStorageProofVotingStrategy.casm.json')
       .toString('ascii'),
   );
   const spaceSierra = json.parse(
--- a/scripts/deploy-space.ts
+++ b/scripts/deploy-space.ts
@@ -1,3 +1,4 @@
+import dotenv from 'dotenv';
 import fs from 'fs';
 import {
   defaultProvider,
@@ -12,6 +13,8 @@ import {
   cairo,
 } from 'starknet';
 
+dotenv.config();
+
 const accountAddress = process.env.ADDRESS || '';
 const accountPk = process.env.PK || '';
 const starknetNetworkUrl = process.env.STARKNET_NETWORK_URL || '';
@@ -21,7 +24,7 @@ async function main() {
   const account = new Account(provider, accountAddress, accountPk);
 
   const l1TokenAddress = '0xd96844c9B21CB6cCf2c236257c7fc703E43BA071'; //OZ token 18 decimals
-  const slotIndex = cairo.uint256(0);
+  const slotIndex = cairo.uint256(8);
 
   const factsRegistryAddress = '0x01b2111317EB693c3EE46633edd45A4876db14A3a53ACDBf4E5166976d8e869d';
   const timestampsRemapperAddress =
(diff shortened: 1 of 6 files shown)


scripts/deploy-space.ts
```diff
@@ -21,7 +24,7 @@ async function main() {
   const account = new Account(provider, accountAddress, accountPk);
 
   const l1TokenAddress = '0xd96844c9B21CB6cCf2c236257c7fc703E43BA071'; //OZ token 18 decimals
-  const slotIndex = cairo.uint256(0);
+  const slotIndex = cairo.uint256(8);
```
pscott: for more clarity, would be nice to detail to what this refers and how we obtained it

scripts/herodotus-interaction-demo.ts
```diff
@@ -30,15 +31,15 @@ async function main() {
...
-    '0x6fa12cffc11ba775ccf99bad7249f06ec5fc605d002716b2f5c7f5561d28081';
+    '0x046ad946f22ac4e14e271f24309f14ac36f0fde92c6831a605813fefa46e0893';
   const evmSlotValueVotingStrategyAddress =
-    '0x06cf32ad42d1c6ee98758b00c6a7c7f293d9efb30f2afea370019a88f8e252be';
+    '0x474edaba6e88a1478d0680bb97f43f01e6a311593ddc496da58d5a7e7a647cf';
 
   const l1TokenAddress = '0xd96844c9B21CB6cCf2c236257c7fc703E43BA071'; //OZ token 18 decimals
-  const slotIndex = 0; // Slot index of the balances mapping in the token contract
-  const voterAddress = '0x2842c82E20ab600F443646e1BC8550B44a513D82';
+  const slotIndex = 8; // Slot index of the checkpoints mapping in the token contract
```
pscott: could be nice to know how one can obtain it (in 1 sentence)

scripts/herodotus-interaction-demo.ts
```diff
@@ -51,9 +52,27 @@ async function main() {
...
   );
+  const numCheckpoints = await l1Token.numCheckpoints(voterAddress);
+  console.log(numCheckpoints);
+
+  // Deriving the keys of the final slot in the checkpoints array for the voter and the next empty slot
+  const checkpointSlotKey =
+    BigInt(
+      ethers.keccak256(
+        ethers.keccak256(
+          `0x${voterAddress.slice(2).padStart(64, '0')}${slotIndex.toString(16).padStart(64, '0')}`,
```
pscott, line 72: so this gives something like `hash(hash(padded_voter_address | padded_slot_index)) + numCheckpoints - 1` ? (where `|` indicates concatenation).

What happens if this operation should overflow? (a hash could be fairly close to uint256::max, and if `numCheckpoints` is big enough, this could happen)

scripts/herodotus-interaction-demo.ts
```diff
@@ -142,8 +161,9 @@ async function main() {
 
   // This is the snapshot L1 block number
   const l1BlockNumber = response.data.path[1].blockNumber;
+  console.log(l1BlockNumber);
```
pscott, line 168: do we wish to keep it ?

scripts/herodotus-interaction-demo.ts
```diff
@@ -212,7 +239,9 @@ async function main() {
           {
             index: '0x0',
             params: CallData.compile({
-              storageProof: storageProofLittleEndianWords64,
+              checkpoint_index: numCheckpoints - BigInt(1),
```
pscott, line 246: is there a world where `numCheckpoints == 0` ? Do checkpoints get updated on every transfer or only on every delegation change?

starknet/src/voting_strategies/evm_slot_value.cairo
```diff
@@ -47,14 +53,40 @@ mod EvmSlotValueVotingStrategy {
...
-            let balance = SingleSlotProof::InternalImpl::get_storage_slot(
-                @state, timestamp, evm_contract_address, slot_index, voter.into(), user_params
+            let checkpoint = SingleSlotProof::InternalImpl::get_storage_slot(
+                @state, timestamp, evm_contract_address, slot_key, checkpoint_mpt_proof
             );
-            balance
+            assert(checkpoint.is_non_zero(), 'Slot is zero');
+
+            // Verify the checkpoint is indeed the final checkpoint by checking the next slot is zero.
+            let state = SingleSlotProof::unsafe_new_contract_state();
```
pscott: can we use the same `state` as before? since it's borrowed earlier I think we can?

starknet/src/voting_strategies/evm_slot_value.cairo
```diff
@@ -91,6 +123,27 @@ mod EvmSlotValueVotingStrategy {
...
+                        .byte_reverse()
+                ]
+                    .span()
+            )
+                .byte_reverse()
+                + integer::U32IntoU256::into(offset)
+        }
+
+        fn decode_checkpoint_slot(slot: u256) -> (u32, u256) {
+            let block_number = slot.low & 0xffffffff;
```
pscott: could be nice to explain why it's on the first 4 bytes, maybe link to the checkpoint struct somewhere?

starknet/src/voting_strategies/evm_slot_value.cairo
```diff
@@ -91,6 +123,27 @@ mod EvmSlotValueVotingStrategy {
...
+            let vp = slot / 0x100000000;
```
pscott: is it better to divide or to `& 0xffff....0000000000` ?

pscott (approved):
Good work. Given this adds a little bit of complexity, it might be good to have some end-to-end testing (maybe by taking hard coded value for tokens / slots etc). I think we would need those for future-proofness and code maintainability !

scripts/herodotus-interaction-demo.ts
bonustrack: On every transfer

Orland0x: it would be zero if they have not delegated their tokens to themself yet. In which case the vp is zero. So would just need a check in sx.js to ensure the voter does indeed have before this call is made (which would just fail otherwise)

starknet/src/voting_strategies/evm_slot_value.cairo
Orland0x: added a detailed comment

starknet/src/voting_strategies/evm_slot_value.cairo
Orland0x: added a detailed comment

starknet/src/voting_strategies/evm_slot_value.cairo
Orland0x: you would then need to shift after by dividing anyway.

starknet/src/voting_strategies/evm_slot_value.cairo
Orland0x: yes probably, will test then update

scripts/herodotus-interaction-demo.ts
Orland0x: removed

scripts/herodotus-interaction-demo.ts
Orland0x: 'a hash could be fairly close to uint256::max' this is quite hard! 

the EVM relies on the rarity of that

PR #610, opened 2024-05-01
feat: oz votes trace 208 voting strategy

Orland0x:
adds a voting strategy to support OZ Votes tokens that use Trace208 for checkpoints, this is the officially supported OZ approach since October 2023

### DIFF
--- a/starknet/src/voting_strategies/oz_votes_trace_208_storage_proof.cairo
+++ b/starknet/src/voting_strategies/oz_votes_trace_208_storage_proof.cairo
@@ -12,8 +12,8 @@ mod OZVotesTrace208StorageProofVotingStrategy {
 
     #[external(v0)]
     impl OZVotesTrace208StorageProofVotingStrategy of IVotingStrategy<ContractState> {
-        /// Returns the delegated voting power of `voter` at the block number corresponding to `timestamp` for tokens that implement OZVotes.
-        ///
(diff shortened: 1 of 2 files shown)


starknet/src/voting_strategies/oz_votes_trace_208_storage_proof.cairo
```diff
@@ -0,0 +1,211 @@
...
+            // https://docs.soliditylang.org/en/v0.8.23/internals/layout_in_storage.html#mappings-and-dynamic-arrays
+            keccak::keccak_u256s_be_inputs(
+                array![
+                    keccak::keccak_u256s_be_inputs(array![mapping_key, slot_index].span())
+                        .byte_reverse()
+                ]
+                    .span()
+            )
+                .byte_reverse()
+                + integer::U32IntoU256::into(offset)
```
pscott, line 126: I guess `offset.into()` doesn't work? Compiler can't infer the type here, based on `keccak_u256s_be_inputs` type signature?

pscott (approved):
Good work, utack :)

starknet/src/voting_strategies/oz_votes_trace_208_storage_proof.cairo
Orland0x: its because we have the generic impl `TIntoU256` which then leads to multiple `Into` impls being in scope simultaneously, hence the need to specify. Kinda annoying.

PR #624, opened 2024-06-27
remove slot is zero check in voting strategies

pscott:
Removes `Slot is zero` check in the voting power strategies, and lets the error be at the space level rather than at the voting strategy level.


bonustrack (approved):
utACK

PR #626, opened 2024-07-03
remove `verifyingContract` from stark sig

pscott:
Removes the `verifyingContract` from the `stark-sig` authenticator. This is done because [SNIP12](https://github.com/starknet-io/SNIPs/blob/main/SNIPS/snip-12.md) does not support the `verifyingContract` key.
We also rename `StarkEIP712` to `SNIP12` for clarity.

The only important changes in this PR are:
1. Removing the `verifyingContract` from and updating the domain hash: https://github.com/snapshot-labs/sx-starknet/blob/4e8ebcec63bb8b6319f927b561c547eacaee6892/starknet/src/utils/constants.cairo#L49-L50 (check git blame)
2. Removing the contract_address from the `get_domain_hash` function (was previously here, you can check git blame: https://github.com/snapshot-labs/sx-starknet/blob/4e8ebcec63bb8b6319f927b561c547eacaee6892/starknet/src/utils/snip12.cairo#L157)

The remaining are `StarkEIP712 -> SNIP12`. I realize now it would've been cleaner with only two distinct commits.

### DIFF
--- a/starknet/src/authenticators/stark_sig.cairo
+++ b/starknet/src/authenticators/stark_sig.cairo
@@ -81,7 +81,7 @@ mod StarkSigAuthenticator {
     use starknet::{ContractAddress, info};
     use sx::interfaces::{ISpaceDispatcher, ISpaceDispatcherTrait};
     use sx::types::{Strategy, IndexedStrategy, UserAddress, Choice};
-    use sx::utils::StarkEIP712;
+    use sx::utils::SNIP12;
 
     #[storage]
     struct Storage {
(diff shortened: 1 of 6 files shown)


Sekhmet:
@pscott could you deploy starkSig authenticator on sepolia with this removed so I could test full flow?

I remember that at some point there was some issue with either Metamask or ArgentX not being happy with verifyingContract not being passed.

pscott:
> @pscott could you deploy starkSig authenticator on sepolia with this removed so I could test full flow?
> 
> I remember that at some point there was some issue with either Metamask or ArgentX not being happy with verifyingContract not being passed.

I think it was metamask that was causing issues, but here it only concerns `sx-sig` so shouldn't be an issue.

New contract: `0x04a7805f0d4b34d801f2fcf7aace7bd990f3acc30e542523998f9c0d5f203055` on sepolia

Sekhmet (approved):
tACK (as https://sepolia.starkscan.co/contract/0x04a7805f0d4b34d801f2fcf7aace7bd990f3acc30e542523998f9c0d5f203055)

PR #634, opened 2024-10-04
Introduce internal components

pscott:
Closes https://github.com/snapshot-labs/workflow/issues/162

In this PR we remove all the `unsafe_new_contract_state()` by making use of [cairo components](https://community.starknet.io/t/cairo-components/101136).

This is mostly just syntax modifications, but it does indeed touch to the precious `.cairo` files.

To make it easy to review, every component was introduced in a different commit.


pscott:
missing: use prefix for storage variables of components to avoid namespace clashing

pscott:
> missing: use prefix for storage variables of components to avoid namespace clashing

done

bonustrack (approved):
utACK

PR #639, opened 2025-06-03
feat: erc721 votes

notV4l:
support erc721 with ERC721VotesVotingStrategy using OZ VotesComponent

(erc20 & erc721 could also share a common OZVotesVotingStrategy)


pscott:
Thanks, we'll use a single Voting Strategy for all `OZVotes` contract in the future :)

: https://github.com/snapshot-labs/sx-starknet/issues/640

PR #641, opened 2026-03-26
refactor: switch to Herodotus satellite

mikjakbiak:


starknet/src/voting_strategies/evm_slot_value.cairo
```diff
@@ -52,26 +51,25 @@ mod EvmSlotValueVotingStrategy {
...
-            let mpt_proof = Serde::<Span<Span<u64>>>::deserialize(ref user_params).unwrap();
 
             // Computes the key of the EVM storage slot from the mapping key and the index of the mapping in storage.
             let slot_key = InternalImpl::get_mapping_slot_key(voter.into(), slot_index);
 
             // Returns the value of the storage slot at the block number corresponding to the given timestamp.
             let slot_value = self
                 .single_slot_proof
-                .get_storage_slot(timestamp, evm_contract_address, slot_key, mpt_proof);
+                .get_storage_slot(timestamp, evm_contract_address, slot_key);
```
beeinger: mpt proofs have to come back, we should use the satellite's verify functions for verification of mpt proofs:
https://github.com/HerodotusDev/satellite/blob/91a5128d309e6f263cfaeef8ffb38b719abaa86f/cairo/src/evm_fact_registry.cairo#L342

proving storage slots directly in mission control + satellite contract is not feasible here

beeinger (changes requested):
Bring back MPT proofs, we should only prove timestamp remapping + storage root & then verify using: https://github.com/HerodotusDev/satellite/blob/91a5128d309e6f263cfaeef8ffb38b719abaa86f/cairo/src/evm_fact_registry.cairo#L342

ethereum/script/DeployDemoToken.s.sol
beeinger, line 1: Maybe call it a DeployDemoToken or something like that? I don't see any indication this is a test/demo if i dont actually go inside and see the line 14

ethereum/script/DeployDemoToken.s.sol
```diff
@@ -0,0 +1,22 @@
...
+import {OZVotesToken} from '../src/token/OZVotesToken.sol';
+
+contract DeployToken is Script {
+  function run() public {
+    uint256 deployerPk = vm.envUint('ETH_PRIVATE_KEY');
+    address deployer = vm.addr(deployerPk);
+
+    vm.startBroadcast(deployerPk);
+
+    OZVotesToken token = new OZVotesToken('Demo Token', 'DEMO', 1_000_000 ether);
```
beeinger, line 14: i mean this

beeinger (changes requested):
Apart from the small naming change everything is solid

beeinger (approved):
Looks good now, ready to merge

pnpm-lock.yaml
```diff
@@ -0,0 +1,7711 @@
+lockfileVersion: '9.0'
```
bonustrack: This repo is using Yarn, we should remove this file from the PR and update yarn.lock instead.

mikjakbiak: fixed

